PR #94, opened 2014-12-11
Use recommended approach for logging.

willscott:
This splits the logging logic into
```logging```, which becomes a client library, wrapping freedom.core().getLogger()
and
```logger```, a freedom module implementing the well known 'console' API, which
can be installed as a custom freedom logger, via
```javascript
freedom('manifest.json', {
  logger: 'logger.json'
}).then(...)
```

In doing so, we can remove the ```core.console``` permission from all other
workers, and the logs will include freedom system errors.

Addresses https://github.com/uProxy/uproxy/issues/503

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/94)


willscott:
I'd prefer to leave ```logging``` as-is, since by doing so and since the external interface of the library hasn't changed, we don't break anyone depending on it in the other repositories.  Happy to change ```logger``` to ```logAggregationModule``` or ```loggingProvider```.  I don't think there is anything called ```freedom['loggers']``` - not sure what that was referring to.  Thoughts / preferences?

iislucas:
@willscott You seem to use `freedom['loggers']` in one of your mocks here:
https://github.com/uProxy/uproxy-lib/pull/94/files#diff-5f0a27f9a7d51fc4f9a9e3bdd3539ac9R10
If this is not a freedom variable, then we probably want to remove it or rename it to something else?

@trevj what do you think of `logging` and `loggingProvider` ?

willscott:
Ah - that's purely for use in that test, freedom is the only global that's
defined, so its simply a property for holding spies that was chosen rather
than creating a new file & custom grunt rule referencing it in order to get
a separate global to reflect that state.

It can be called anything.

willscott:
Okay to change name to loggingProvider?  (@trevj)

This pull request has been out for a month now, and I'd like to not have to refactor it through too many other changes.

src/logging/logging.ts
```diff
@@ -65,128 +27,61 @@ module Logging {
...
+        realLog = logger.debug;
+      } else if (level === 'info') {
+        realLog = logger.info;
+      } else if (level === 'warn') {
+        realLog = logger.warn;
+      } else if (level === 'error') {
+        realLog = logger.error;
       }
+      
+      return realLog(formatStringMessageWithArgs_(msg, args));
```
trevj: Ooh, ugly. Use an else block.

src/logging/logging.ts
```diff
@@ -65,128 +27,61 @@ module Logging {
...
+        doLog(msg.level, tag, msg.msg);
+      }
+      delete waiters[tag];
     }
   }
 
   export class Log {
-    constructor(private tag_:string) {}
+    constructor(private tag_:string) {
+      freedom['core']().getLogger(this.tag_).then((logger:freedom.Logger) => {
```
trevj: A member variable of type Promise.

It's arguable how understandable the `waiters` array is but it's definitely more complex than introducing a member variable to `Log`.

willscott:
Thanks for looking over the code again @trevj - can you also weigh in on the renaming decision @iislucas has asked for your input on in this thread?

trevj:
`loggingProvider` sounds good to me. I was having a hard distinguishing `logger` from `logging`!

src/logger/logger.d.ts
```diff
@@ -0,0 +1,35 @@
...
+  interface Message {
+    timestamp :Date; // the timestamp the log was called (in core runtime).
+    level :string; // one of D=Debug, I=Info, W=Warning, E=Error
+    tag :string; // any string, used for viewing specific module logs.
+    message :string; // the actual log message.
+  }
+
+  // This is the real internal interface available to other core modules/the
+  // top level environment.
+  function makeMessage(level:string, tag:string, msg:string)
```
trevj: A secondary API could work, although I'm not familiar with the nuts and bolts of doing it. Can you update one of the sample apps to call `setConsoleFilter`? I'd be happy to then update uproxy-networking.

willscott: can this happen in a second pull request? It should ideally be done with getLogs, and it's a whole other API that we'll want to come to consensus on.

willscott:
Rename made.
Refactor of logging to use promise rather than waiters per @trevj  done.
PTAL

trevj:
Thanks for renaming and removing `waiters`. I like it a lot better.

Regarding `setConsoleFilter` and API discussion and followup pull request, I just don't want to leave us without the ability to run the networking samples with debug logging (is that you meant? your reply doesn't appear in the Github interface, so I'm guessing the context)

willscott:
Logging defaults to debug level logging, at least from how i understand https://github.com/uProxy/uproxy-lib/blob/freedom-logger/src/loggingprovider/loggingprovider.ts#L33 so that shouldn't be a problem?

We don't need to publish an NPM release of uproxy-lib at this stage. I'd just like to have this be a separate pull request, since it feels like there's another round of design to do there. I'd prefer to not have to worry about this change overlapping with things like the ```require``` refactor, which I don't fully understand the consequences of.

src/loggingprovider/loggingprovider.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "uProxy Logging Provider",
+  "description": "Logging Provider for aggregating uProxy logs, and filtering.",
+  "app": {
+    "script": "loggingProvider.js"
```
trevj: `loggingprovider.js`

trevj:
OK, I think I more or less have the gist of things here.

Two things I would like to have before merging this:
 1. Please test the Simple Freedom Chat sample app! I just patched it in and there's a typo in the logging provider manifest which prevents *any* logging being output. Additionally, having fixed that typo in my client, I'm still not seeing any debug logging. It seems calls to `console.debug` inside loggingprovider.js have no effect -- there should be one for each signalling channel message.
 1. Let's make whatever changes are needed to the loggingprovider manifest to provide access to the `setConsoleFilter` call, and call it from Simple Freedom Chat. I suspect that's very easy for you and much harder for me. I'm happy to later update uproxy-networking.

willscott:
PTAL

willscott:
Debugging wasn't being shown because the ```simple-freedom-chat``` told freedom to only log at the 'log' and above level (https://github.com/uProxy/uproxy-lib/blob/dev/src/samples/simple-freedom-chat/main.ts#L12)

trevj:
There's a bunch of calls to `Logging.setConsoleFilter` so I'm afraid this is a breaking change. Let's bump the major version number then merge.

willscott:
Version bumped. Merging.  When you get a chance, give it a push.

PR #100, opened 2015-01-07
Big refactor to use require-style typescript + many cleanups

iislucas:
Addresses: https://github.com/uProxy/uproxy/issues/363
This is a significant code re-organization & rationalization of our build process.

Observations/Remarks: 
- This is rather a lot... sorry for the huge pull request! Hopefully no such big refactors of our build system  for a while now. 
- Using require in typescript was surprisingly easy (only had to spend 3 hours). 
- It makes dependencies more explicit and easier to follow
- The browserify stuff needed to actually use require statements in JS and manage building everything etc was surprisingly difficult to figure out (it interacted with some extra needed understanding of freedom, fixes to tests, and refinement of structure/location of files) (2 weeks)
- I think this is a LOT nicer, it's a now build structure with very few hacks (just one one's needed for using grunt).

Summary of work: 
- All typescript files now use require for importing typescript dependencies (one's which result in a require statement in the output code that is then linked by browserify)
- Separates interfaces for consumer/producer of handler-queues (implements: https://github.com/uProxy/uproxy/issues/657).
- Avoids test mock freedom in the for logging test (instead tests the interaction with freedom using freedom's APIs)
- Separates the `freedom-module-env` from the `freedom-core-env` typescript definitions. These are now two different definitions. This means that compilation of a typescript file for the `freedom-core-env` must now be done separately from a file that references  `freedom-module-env` (at least until freedom names the objects differently).
- Removed the need for writing/maintaining `*.d.ts` files. Now, `*.d.ts` files are now used only in the output `dev` and `dist` directories, and are generated programatically. 
- We now compile directly from `src` directory into `build/dev`. No more managing sym-links of typescript files in the build directory. 
- There is now a single build directory (like the main uProxy) which has `dev` and `dist` as subdirectories (no more root level dist). 
- Uses browserify as a JS linker, which is aware of the typescript compiled `require` statements. We no longer need to maintain the right web-page `<script src=...>` inclusions, nor chrome app script references, nor multiple freedom module script references: you now only need the single top level reference to the JS file. 
  - Dependencies used by sample apps are minimized: each freedom module is browserified into a single JS file: it contains only the files it uses, and is referenced from the freedom json manifest file. 
  - For spec-files, freedom-modules, and top-level sample-app files, each file needs to go through a browserification process. Browserify produces a `*.static.js` file that has all dependencies included. This automatically pulls in all the required files. 
  - Typescript compilation freedom modules must now be separately from compilation of a file in the core environment because they have different types for the (different, but named the same) `freedom` objects. 
- Move to using a new `install_dev_deps.sh` script: Given that tsd needs to be run to setup typescript definitions, and npm install scripts can't be selective for dev-dependencies (which means that you don't want to use an npm scripts for this), I've added in a little bash script for installing dev dependencies (it runs both the npm install, the appropriate tsd command, and tsc for build tools used by the gruntfile).

TESTED: 
- grunt test
- Setup on drone.io, shippable, and travis
- Ran sample apps & interactively tested them

TODO (in followup pull requests): 
- Make sure that uproxy-networking can use it: that means require-style version of the uproxy-networking repo.
- update coding guide & document some of the gotacha's
- Support sourcemaps
- Experiment with gulp instead of grunt.
- Think of a better name for some of the freedom typings & move freedom typings back to freedom-typescript-api repo

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/100)


ryscheng:
Cool! :)

trevj:
This seems very cool.

I'm surprised to see it reduces the level of indentation, e.g. `arraybuffers.ts`, by eliminating the need for a TypeScript module. That's very nice.

Compilation directly from `src/` sounds very appealing. I'm very curious to see the updated Gruntfile.

I think trying to compile uproxy-networking against this will tease out a bunch of issues, if they exist. Awesome.

iislucas:
Eventually, I think I'd like to move to gulp instead of grunt, but that should wait or another pull request I think (already 13k line changes in this pull request!). Anyway, the updated gruntfile is here for your perusal: https://github.com/uProxy/uproxy-lib/blob/iislucas/Gruntfile.coffee  :)

willscott:
Is this pull-request ready for review, or just the gruntfile?

iislucas:
Things still to be changed in this pull request before real review: 
* Support exporting of .d.ts files in a distribution directory. 
* Try to fix grunt-ts weirdness in terms of output directory construction (maybe have to go back to horrible sym-linking).
* Make sure that uproxy-networking can use it reasonably: that means require-style version of uproxy0-networking.
* Write summary of everything that was done & document some of the gotacha's
* Write how it was tested.

iislucas:
@trevj and others, probably good to start reviewing this now; I think it's just about ready :)

Gruntfile.coffee
```diff
@@ -1,108 +1,286 @@
...
+
+# Makes the base development build, excludes sample apps.
+taskManager.add 'base-dev', [
+  'copy:dev'
+  'ts:dev'
+  'browserify:loggingProvider'
+]
+
+# Makes the development build, includes sample apps.
+taskManager.add 'dev', [
```
willscott: this is the same task as sample, if I'm reading definitions correctly. consider condensing to 1 name?

package.json
```diff
@@ -17,28 +17,30 @@
...
+    "grunt-ts": "^2.0.1",
+    "grunt-tsd": "^0.1.0",
+    "tsify": "^0.7.1",
+    "typescript": "^1.3.0"
   },
   "scripts": {
-    "test": "grunt test",
-    "prepublish": "grunt test"
+    "test": "grunt unit_tests",
+    "postinstall": "grunt setup"
```
willscott: i think this doesn't do what you want - i believe it runs every time your module is installed, even as a dependency of something else - in which case the devDependencies won't exist, so grunt won't be able to run setup for you.

see: https://docs.npmjs.com/misc/scripts#note-install-scripts-are-an-antipattern

src/build-tools/common-grunt-rules.ts
```diff
@@ -0,0 +1,103 @@
...
+    return {
+      src: [ path.join(this.config.devBuildDir, filepath + '.js')],
+      dest: path.join(this.config.devBuildDir, filepath + '.static.js'),
+      options: {
+        debug: true,
+      }
+    };
+  }
+
+  // Grunt copy target creator: for copying freedom.js to
```
willscott: grammar

src/freedom/typings/freedom-core-env.d.ts
```diff
@@ -0,0 +1,17 @@
...
+// The freedom file, when loaded, will assume that a new variable |freedomcfg|
+// on the global window object may be set.
+interface Window {
+  // The freedom config variable can be set in the window to let an application
+  // register additional core-providers. When Freedom starts up, if the
+  // |freedomcfg| var is defined, it will call it passing it the internal core-
+  // provider registration function. |providerName| should be a name of a core-
+  // provider as defined in |freedom/interface/core.js|, and the function-
+  // argument should be a class that meets that interface of |providerName|.
+  freedomcfg ?: (register:(providerName:string, classFn:Function) => void)
```
willscott: this hasn't been true for a long time, and can be removed?

src/loggingprovider/loggingprovider.i.d.ts
```diff
@@ -0,0 +1,60 @@
...
+  // Logs message in warn level.
+  warn(source:string, msg: string) : void;
+  // Logs message in error level.
+  error(source:string, msg: string) : void;
+}
+
+// Interface for managinge & retreiving log messages.
+// Note: this is really a fake class: all data is in fact global.
+// TODO: rename this to LoggingManager or something sensible.
+export interface Controller {
```
willscott: Resolve TODO above

tools/common-grunt-rules.js
```diff
@@ -1,106 +1,66 @@
...
+            dest: path.join(this.config.devBuildDir, filepath + '.static.js'),
             options: {
-                sourceRoot: 'build/',
-                target: 'es5',
-                comments: false,
-                noImplicitAny: true,
-                sourceMap: true,
-                declaration: false,
-                fast: 'always'
+                debug: true,
```
willscott: this leaves source-map in the final file. may want to use `exorcist` to separate sourcemap to separate file.

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)



---

OK, so, I think this represents a huge improvement.

I have a whole bunch of comments, many of which are just relating to typos and leftover bits and bobs which are inevitable in a pull request of this size.

For source, I pretty much just set my diff tool to ignore changes in leading whitespace and assume you've only changed the indentation, imports, and copied comments across from `.d.ts` files (so happy to see these mostly disappear, BTW).

Can you merge in the recent changes to dev? This affects `package.json` in particular, which I have ignored for now.

---

<sup>**[.gitignore, line 15 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw17t90Bs3WfHQNP18)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/.gitignore#L15)):</sup>
What creates this folder?

---

<sup>**[Gruntfile.coffee, line 6 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgv_IIaMX5LsJSSZcal)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L6)):</sup>
It seems like the motivation behind `TaskManager` has diminished a lot with this pull request?

---

<sup>**[Gruntfile.coffee, line 11 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvYXkhj7bzw_dmThaM)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L11)):</sup>
Can you think of a better name than 'setup'?

---

<sup>**[Gruntfile.coffee, line 29 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgv_XtB5PmkFFYUpUSw)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L29)):</sup>
I prefer the current name of `build`, which is also a well recognized convention.

---

<sup>**[Gruntfile.coffee, line 69 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw2qp5V4MgKGO5bkiJ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L69)):</sup>
Can you add a target called 'test' which depends on this?

---

<sup>**[Gruntfile.coffee, line 70 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgva-cBw7mVKLLmw_Ak)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L70)):</sup>
I think we can avoid spelling out each target by just listing:
```
'dev'
'browserify'
'jasmine'
```

---

<sup>**[Gruntfile.coffee, line 96 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvaKyXgg1MC9A0xJ3P)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L96)):</sup>
Is this used?

---

<sup>**[Gruntfile.coffee, line 100 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw1U9OTOr7pfhE_K9l)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L100)):</sup>
It's not obvious why you copy non-TypeScript files into `build/dev`.

Is it purely so we can later copy them from `build/dev` -> `build/dist`? If so, I'd suggest that, for a component xxx, `copy:dist` have two steps:
 1. copy everything from `src/xxx` into `build/dist/xxx`
 1. copy everything from `build/dev/xxx` into `build/dist/xxx`

---

<sup>**[Gruntfile.coffee, line 123 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw1myaD9T1BKE0jBGO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L123)):</sup>
After having built my clone, I can't find any `*.dynamic.js` files inside.

---

<sup>**[Gruntfile.coffee, line 134 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw39W0RAS6RGcWL-Vs)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L134)):</sup>
When it comes to uproxy-networking, will we have to copy each dependency separately?

---

<sup>**[Gruntfile.coffee, line 146 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw2Pg5YXlwQ4-fql19)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L146)):</sup>
No need for this comment; the `tools` target lists it as a sub-target.

---

<sup>**[Gruntfile.coffee, line 163 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvbZqdbE7fG7PQSbON)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L163)):</sup>
How about `install` and `update` as target names?

---

<sup>**[Gruntfile.coffee, line 182 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgvz3CfejwYHfm22ljA)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L182)):</sup>
It's really neat that we can compile a bunch of folders with one command.

However, two things:
 1. It's not obvious why we compile sample apps separately. In fact, we list each sample app twice! This will get really boring in uproxy-networking. Is it just to avoid generating declarations?
 1. If these really have be listed individually, we definitely need a re-useable rule in `common-grunt-rules`!

---

<sup>**[Gruntfile.coffee, line 252 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw4Wavg4U3DKycxhna)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L252)):</sup>
I'm trying to figure out how this path could be correct. The only copy in me repo is at `src/freedom/mocks/`. Are you sure this is actually loaded?

---

<sup>**[Gruntfile.coffee, line 258 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw0XTO_2AcK6raG-Fm)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L258)):</sup>
So, each browserify target results in a `.static.js`?

---

<sup>**[Gruntfile.coffee, line 270 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvbHP58lA2pDy8Pg1N)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L270)):</sup>
This comment seems out of place.

---

<sup>**[Gruntfile.coffee, line 277 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw0f90wGDaSWW7tv24)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L277)):</sup>
remove commented code

---

<sup>**[README.md, line 22 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgrcFfUF8-nqbdnEsqm)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/README.md#L22)):</sup>
Similar to what you'd done for `tsd`, I think we an eliminate the need to commit `taskmanager.js` and `common-grunt-rules.js` by pushing their compilation to some sort of base target.

No need to do this now but how about we create an issue?

---

<sup>**[README.md, line 28 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgrcuCeGjui69Yup4ZI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/README.md#L28)):</sup>
typo: freeom -> freedom

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 8 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvSq2dNZ6zbu_62aOu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L8)):</sup>
So these are great things to have for tests but regarding naming I think we can do better!

These objects are usually called `AbstractWhatever` in Java-land or, less confusingly and probably more appropriately for JavaScript-land, `SkeletonWhatever`. What do you think?

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 112 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvTUpggrrxe49zioCo)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L112)):</sup>
Should the return type by `any`?

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 124 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvTaxccDm7ieRW2quI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L124)):</sup>
remove commented lines

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 148 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvTezr2JFcSZI7Hs5-)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L148)):</sup>
remove commented lines!

---

<sup>**[src/freedom/sample-code/freedom-core-env.spec.dynamic.ts, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvTnc6KStsOpvAH6KQ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/sample-code/freedom-core-env.spec.dynamic.ts#L1)):</sup>
```js
// This is a dummy file to make sure that we typecheck the
// freedom-typescript-api files.
```

---

<sup>**[src/freedom/sample-code/freedom-core-env.spec.dynamic.ts, line 10 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvTzg-Ds4xO0J6Ip96)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/sample-code/freedom-core-env.spec.dynamic.ts#L10)):</sup>
indentation

---

<sup>**[src/freedom/typings/freedom-common.d.ts, line 26 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgrdPbDk_X-DdgRht94)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/typings/freedom-common.d.ts#L26)):</sup>
Not saying I have a better suggestion but maybe we can think of a better name? :-)

---

<sup>**[src/freedom/typings/freedom-common.d.ts, line 113 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvUi4Kf1gRlM1ot48d)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/typings/freedom-common.d.ts#L113)):</sup>
You just blew my mind.

Does this conflict with the `module freedom {` declaration at the top of this file? Perhaps it would be better to rename that?

---

<sup>**[src/handler/aggregate.ts, line 11 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgrdmm9lC_HypuT9rCM)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/handler/aggregate.ts#L11)):</sup>
indentation

---

<sup>**[src/handler/events.ts, line 3 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgre1patYW_xrEwHNB0)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/handler/events.ts#L3)):</sup>
Several typos in this paragraph.

---

<sup>**[src/handler/events.ts, line 7 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgre9X8vyBuUDHj0udC)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/handler/events.ts#L7)):</sup>
Does this comment help? :-)

---

<sup>**[src/handler/queue.ts, line 3 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgreOqKSOHinx2QNUWr)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/handler/queue.ts#L3)):</sup>
Several typos in this paragraph.

---

<sup>**[src/loggingprovider/loggingprovider.json, line 16 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvVKsu4_vIwPyJIZtU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/loggingprovider/loggingprovider.json#L16)):</sup>
If we're going to change the name, let's change throughout. I'd suggest renaming in a separate pull request.

---

<sup>**[src/samples/simple-freedom-chat/messages.i.d.ts, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvVv4OIUqF1XTO7kae)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/samples/simple-freedom-chat/messages.i.d.ts#L1)):</sup>
So you've used the `*.i.d.ts` in some places, like here, but not in others, e.g. `webrtc/types.d.ts`.

Personally, I don't think the `.i.` is necessary, or pretty.

---

<sup>**[src/webrtc/types.d.ts, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvWINb0i-Dja3AuIq1)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/webrtc/types.d.ts#L1)):</sup>
Can you think of a better name than `types.ts`?

---

<sup>**[src/webrtc/webrtc.ts, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvWFqxQexijn6LrI0H)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/webrtc/webrtc.ts#L1)):</sup>
Why this file on its own?

---

src/freedom/typings/freedom-core-env.d.ts
iislucas: removed. Has anything replaced it?

src/build-tools/common-grunt-rules.ts
iislucas: tweaked.

package.json
iislucas: hrm... no good way in npm to run installers for non-npm dependencies by the sound of it... I'll write a bash script for the install process so there's only a single point that needs to be extended as we move forward.

Gruntfile.coffee
iislucas: Yes, well spotted. Merged into just `dev`.

src/loggingprovider/loggingprovider.i.d.ts
iislucas: done.

tools/common-grunt-rules.js
iislucas: Disabled sourcemaps for now. This pull request is already big enough :)  I'll do sourcemaps in a followup.

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)







---

<sup>**[.gitignore, line 15 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw17t90Bs3WfHQNP18)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/.gitignore#L15)):</sup>
WebStorm editor which has a nice TypeScript mode. 

---

<sup>**[Gruntfile.coffee, line 6 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgv_IIaMX5LsJSSZcal)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L6)):</sup>
Yes, a little. grunt still does no dependency resolution, and doesn't check for cycles. So TaskManager is still being a little helpful.  

---

<sup>**[Gruntfile.coffee, line 11 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvYXkhj7bzw_dmThaM)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L11)):</sup>
`install_typescript_definitions` ?

---

<sup>**[Gruntfile.coffee, line 29 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgv_XtB5PmkFFYUpUSw)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L29)):</sup>
how do you think we should distinguish between the dev build and dist build? call them `build` and `dist`? If you like I could provide another alias for `dev` called `build`?

---

<sup>**[Gruntfile.coffee, line 69 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw2qp5V4MgKGO5bkiJ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L69)):</sup>
sure; you mean so that the default `test` task runs unit_tests? done.

---

<sup>**[Gruntfile.coffee, line 70 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgva-cBw7mVKLLmw_Ak)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L70)):</sup>
There are browserify tasks that are not part of unit tests... but you're right the jasmine tasks are... but I kind of prefer to see the jasmine and broswerify together here as it shows the relationship. I don't feel super strongly, but just a slight preference. 

---

<sup>**[Gruntfile.coffee, line 96 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvaKyXgg1MC9A0xJ3P)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L96)):</sup>
It's really just a reminder for templates based on the package file... unused. I can remove it if you think that's better, but I think it's standard stuff in a gruntfile. Hoping to nuke gruntfile in favour of gump one day.

---

<sup>**[Gruntfile.coffee, line 100 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw1U9OTOr7pfhE_K9l)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L100)):</sup>
To make the dev build usable from the dev directory, you need the json/js/html etc to be there too. The plan is that `dist` will end up being uglified in some way; dev will be a development build with readable source. Make sense?

---

<sup>**[Gruntfile.coffee, line 123 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw1myaD9T1BKE0jBGO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L123)):</sup>
There's two typescript files in the `src/freedom/` dir that are intended for testing freedom, but which we don't want to try and compile with everything else because they each assume a different freedom context. I'm using `*.dynamic.js` for files that must be compiled/browserified independently of other files (added to pull request description, and I'll make sure to document it in the coding guide when I update that too). 

---

<sup>**[Gruntfile.coffee, line 134 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw39W0RAS6RGcWL-Vs)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L134)):</sup>
I think we'll probably write a little helper function so that we can write them in a list. 

---

<sup>**[Gruntfile.coffee, line 146 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw2Pg5YXlwQ4-fql19)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L146)):</sup>
removed. 

---

<sup>**[Gruntfile.coffee, line 163 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvbZqdbE7fG7PQSbON)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L163)):</sup>
Will pointed out that npm install scripts are always run, and generally discouraged, so I've removed grunt-tsd and I am using just the tsd command from a bash script. I added a bash script for the installation of all dev dependencies. 

---

<sup>**[Gruntfile.coffee, line 182 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgvz3CfejwYHfm22ljA)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L182)):</sup>
We need to run typescript for the sample apps twice because they each compilation has a different javascript context (the freedom objects are different). This can be fixed in two ways I think: we could name the freedom object different things in different contexts, e.g. `freedomCoreEnv` vs `freedomModuleEnv`. Or we can run separate compilations for each JS runtime environment (this is what we do ). 

I agree, that this should (eventually) be done more automatically. I'll look at that when I update uproxy-networking and have a few more examples; or I might punt on it for moving to gulp. Grunt is really bad for having it's notion of build upside down. 

---

<sup>**[Gruntfile.coffee, line 252 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw4Wavg4U3DKycxhna)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L252)):</sup>
well spotted; that is left over from a different way I tried to make tests work. Removed.

---

<sup>**[Gruntfile.coffee, line 258 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw0XTO_2AcK6raG-Fm)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L258)):</sup>
exactly! Each runtime environment now only needs a single top-level file, and that file specifies explicitly its dependencies using require statements. 

---

<sup>**[Gruntfile.coffee, line 270 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvbHP58lA2pDy8Pg1N)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L270)):</sup>
oops. removed. 

---

<sup>**[Gruntfile.coffee, line 277 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw0f90wGDaSWW7tv24)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L277)):</sup>
done.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)



---

OK, I think I've addressed everything. Thanks for the great comments. I'm much happier with it already. :) PTAL. 

---

<sup>**[README.md, line 22 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgrcFfUF8-nqbdnEsqm)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/README.md#L22)):</sup>
Neat idea! (That wasn't possible when it was a grunt-rule, because grunt needed task-manager. But now I've moved the installation of developer dependencies to a shell script, I can do this...) Done: this also removes the `tools` tasks, nice one.  

---

<sup>**[README.md, line 28 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgrcuCeGjui69Yup4ZI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/README.md#L28)):</sup>
fixed. 

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 8 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvSq2dNZ6zbu_62aOu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L8)):</sup>
Don't mind much; renamed things with errors to Abstract... and things with nulls to skeletons.

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 112 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvTUpggrrxe49zioCo)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L112)):</sup>
Good point, improved style.

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 124 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvTaxccDm7ieRW2quI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L124)):</sup>
done

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 148 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvTezr2JFcSZI7Hs5-)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L148)):</sup>
done.

---

<sup>**[src/freedom/sample-code/freedom-core-env.spec.dynamic.ts, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvTnc6KStsOpvAH6KQ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/sample-code/freedom-core-env.spec.dynamic.ts#L1)):</sup>
done.

---

<sup>**[src/freedom/sample-code/freedom-core-env.spec.dynamic.ts, line 10 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvTzg-Ds4xO0J6Ip96)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/sample-code/freedom-core-env.spec.dynamic.ts#L10)):</sup>
done

---

<sup>**[src/freedom/typings/freedom-common.d.ts, line 113 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvUi4Kf1gRlM1ot48d)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/typings/freedom-common.d.ts#L113)):</sup>
No conflict. This is a typescript thing: by declaring a module using the quoted style "freedom.i", it allows you to write `import freedoTypes = require("freedom.i")`. 

I've renamed `freedom.i` to `freedom.types` and added a lengthier comment to explain the magic. :) 

---

<sup>**[src/handler/aggregate.ts, line 11 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgrdmm9lC_HypuT9rCM)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/handler/aggregate.ts#L11)):</sup>
done.

---

<sup>**[src/handler/events.ts, line 3 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgre1patYW_xrEwHNB0)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/handler/events.ts#L3)):</sup>
fixed.

---

<sup>**[src/handler/events.ts, line 7 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgre9X8vyBuUDHj0udC)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/handler/events.ts#L7)):</sup>
:) removed.

---

<sup>**[src/handler/queue.ts, line 3 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgreOqKSOHinx2QNUWr)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/handler/queue.ts#L3)):</sup>
clarified/improved.

---

<sup>**[src/loggingprovider/loggingprovider.json, line 16 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvVKsu4_vIwPyJIZtU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/loggingprovider/loggingprovider.json#L16)):</sup>
There used to be two different things here that has the same name. Naming is now consistent everywhere (I think). But you are right, that it would have maybe been better to separate the renaming, but as I was confused during the cleanup, so I renamed on route. `loggingcontroller` is the name of one of the APIs of the loggingprovider module. I don't want to disentangle this again so I'd rather leave this all together. Happy to do an in person review/discussion if helpful. 

---

<sup>**[src/samples/simple-freedom-chat/messages.i.d.ts, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvVv4OIUqF1XTO7kae)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/samples/simple-freedom-chat/messages.i.d.ts#L1)):</sup>
Good point; now using `.types.d.ts` for files that contain only types, and `.enums.ts` for files that contain only enums. No more `.i.d.ts`

---

<sup>**[src/webrtc/types.d.ts, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvWINb0i-Dja3AuIq1)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/webrtc/types.d.ts#L1)):</sup>
Done: `webrtc.types.d.ts` ?

---

<sup>**[src/webrtc/webrtc.ts, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvWFqxQexijn6LrI0H)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/webrtc/webrtc.ts#L1)):</sup>
Users only really need to know the peerconnection interface, and the factory constructor. This allows us to rename the interface to `Transport` later, and have WebRtc as just one instance of a how to do transport. Making steps towards supporting web-sockets/other transport providers. 

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)



---

Very close now. Thanks for addressing the comments!

---

<sup>**[.travis.yml, line 7 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhA01Gob0IToatF-kks)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/.travis.yml#L7)):</sup>
No need for this since you've already configured `npm test` to invoke `grunt unit_tests` (`npm test` is Travis' default command).

---

<sup>**[Gruntfile.coffee, line 7 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvYXkhj7bzw_dmThaM)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L7)):</sup>
Ugh, no :-)

But anyway it looks like this task disappeared, in lieu of the shell script?

---

<sup>**[Gruntfile.coffee, line 16 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgv_XtB5PmkFFYUpUSw)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L16)):</sup>
It's not that I feel very strongly about this -- just that I thought `build` was a fine name.

`build` and `dist` sound good.

---

<sup>**[Gruntfile.coffee, line 51 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw2qp5V4MgKGO5bkiJ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L51)):</sup>
Great. And hopefully  `integration_tests` in future.

---

<sup>**[Gruntfile.coffee, line 52 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgva-cBw7mVKLLmw_Ak)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L52)):</sup>
I don't feel very strongly about it.

Just pointing out that it will get tiresome listing every target and may create unnecessary friction when writing tests. Up to you.

---

<sup>**[Gruntfile.coffee, line 81 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvaKyXgg1MC9A0xJ3P)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L81)):</sup>
I didn't know that. Thanks.

---

<sup>**[Gruntfile.coffee, line 85 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw1U9OTOr7pfhE_K9l)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L85)):</sup>
Ah, I see. That seems reasonable.

In that case, I would suggest just copying all non-TypeScript files, e.g.
```
src: ['**/*', '~**/*.ts']
```

---

<sup>**[Gruntfile.coffee, line 103 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAMPb7UIoNFsz1ZY1w)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/Gruntfile.coffee#L103)):</sup>
Again, why not just:
```
src: ['**/*', '~**/*.spec.dynamic.js']
```

---

<sup>**[Gruntfile.coffee, line 108 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw1myaD9T1BKE0jBGO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L108)):</sup>
OK, I think I follow but it's confusing because there aren't any `.dynamic.ts` files yet.

Excluding specs from `dist` sounds right.

---

<sup>**[Gruntfile.coffee, line 119 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw39W0RAS6RGcWL-Vs)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L119)):</sup>
Wait, won't browserify create a `.static.js` for each sample app's freedomjs module? Maybe we don't need a general solution? Will it be enough to create a rule that copies freedomjs + logger thing to a sample app's directory?

---

<sup>**[Gruntfile.coffee, line 126 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvbZqdbE7fG7PQSbON)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L126)):</sup>
I don't have a good argument against but don't forget that we had a `setup.sh` in the uproxy repo which we ended up discarding in favour of simply documenting the steps in the `README`.

---

<sup>**[Gruntfile.coffee, line 131 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgvz3CfejwYHfm22ljA)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L131)):</sup>
I see.

This seems like something we will be strongly motivated enough by to not to forget doing, so okay :-)

---

<sup>**[Gruntfile.coffee, line 205 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw0XTO_2AcK6raG-Fm)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L205)):</sup>
Awesome.

---

<sup>**[install_dev_deps.sh, line 1 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAB0qH_ezDD26ZaRps)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/install_dev_deps.sh#L1)):</sup>
That's quite a name! `setup.sh`?

---

<sup>**[install_dev_deps.sh, line 11 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhA1848M__ExZnUXv8S)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/install_dev_deps.sh#L11)):</sup>
Personally, I wouldn't bother printing the statements.

`bash -v` will achieve the same thing for those users who need to know what the script does (hopefully nobody, right, if the script is well written? :-))

---

<sup>**[install_dev_deps.sh, line 20 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAAcwIqqyEYGoElv66)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/install_dev_deps.sh#L20)):</sup>
`pushd` and `popd` will allow your script return the shell to the original working directory when the script terminates successfully.

```bash
pushd > /dev/null
# do stuff
popd > /dev/null
```

---

<sup>**[install_dev_deps.sh, line 30 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAAxnLA8dE-GqSHNjv)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/install_dev_deps.sh#L30)):</sup>
I assume the various commands here will complain if there are errors so I would personally omit this.

---

<sup>**[README.md, line 21 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAd1OcGqu40TVzmXtF)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/README.md#L21)):</sup>
You can just put these on one line.

---

<sup>**[README.md, line 34 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAd5C5bUmAwp7tsGUy)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/README.md#L34)):</sup>
Don't forget typings.

---

<sup>**[README.md, line 53 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgrcFfUF8-nqbdnEsqm)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/README.md#L53)):</sup>
YES!

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 8 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvSq2dNZ6zbu_62aOu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L8)):</sup>
You know, I didn't even spot that you had some of these throwing Error and some not.

If we really need both, the naming seems reasonable.

However, it's not clear to me that there's much overlap between what I see as the two main cases:
 1. objects that throw `Error` highlight to the test writer when they have omitted to implement an essential method
 1. objects that `return null` free the test writer from repeatedly mocking frequently called methods which are useless in a test unit environment, e.g. `getLogger`

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 10 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAhZOojBhcIDaCINKf)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L10)):</sup>
nit: it's not a mock!

---

<sup>**[src/freedom/typings/freedom-common.d.ts, line 26 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgrdPbDk_X-DdgRht94)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/typings/freedom-common.d.ts#L26)):</sup>
ping

(code smell: naming something "thing")

---

<sup>**[src/handler/queue.ts, line 3 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgreOqKSOHinx2QNUWr)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/handler/queue.ts#L3)):</sup>
Wow! Cool.

---

<sup>**[src/logging/logging.ts, line 5 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAPg4zC5buTymFX8Y1)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/src/logging/logging.ts#L5)):</sup>
remove commented code

---

<sup>**[src/loggingprovider/loggingprovider.json, line 16 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvVKsu4_vIwPyJIZtU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/loggingprovider/loggingprovider.json#L16)):</sup>
Oh...I get it. This logging stuff is pretty complex. OK. `controller` might be a better name but I really only care about containing the craziness, which it seems to be.

---

<sup>**[src/webrtc/webrtc.ts, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvWFqxQexijn6LrI0H)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/webrtc/webrtc.ts#L1)):</sup>
While, that's something we want to do at some point I don't think this helps towards it in any significant way.

I suggest we take advantage of automatic the `.d.ts` generation as we do everywhere else until we've a good reason not to.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)



---

PTAL. 

Just a couple of small things to discuss. 

Also did a couple of small things: renamed a freedom json file to `freedom-module.json` and merged with head of `dev`.

---

<sup>**[Gruntfile.coffee, line 7 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvYXkhj7bzw_dmThaM)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L7)):</sup>
yes. :) 

---

<sup>**[Gruntfile.coffee, line 16 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgv_XtB5PmkFFYUpUSw)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L16)):</sup>
ACK.

---

<sup>**[Gruntfile.coffee, line 52 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgva-cBw7mVKLLmw_Ak)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L52)):</sup>
Yeah, hoping to clean this up later as a move to using gulp. 

---

<sup>**[Gruntfile.coffee, line 85 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw1U9OTOr7pfhE_K9l)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L85)):</sup>
good point! done. 

---

<sup>**[Gruntfile.coffee, line 103 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAMPb7UIoNFsz1ZY1w)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/Gruntfile.coffee#L103)):</sup>
Only because I wasn't smart enough to think of it. :) done. 

---

<sup>**[Gruntfile.coffee, line 108 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw1myaD9T1BKE0jBGO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L108)):</sup>
tweaked. 

---

<sup>**[Gruntfile.coffee, line 119 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-Jgw39W0RAS6RGcWL-Vs)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L119)):</sup>
Yes, each module needs to fully specify it's dependencies. We currently include all deps by listing them in the json file and making sure that all paths line up. We get a slightly smoother process by compiling with typescript files with require, and then letting browserify link all the files together for us. 

I've created a more systematic solution as part of a followup that will go along with uproxy-networking, see here: 

---

<sup>**[Gruntfile.coffee, line 153 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvbZqdbE7fG7PQSbON)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/Gruntfile.coffee#L153)):</sup>
ACK

---

<sup>**[install_dev_deps.sh, line 1 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAB0qH_ezDD26ZaRps)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/install_dev_deps.sh#L1)):</sup>
done.

---

<sup>**[install_dev_deps.sh, line 11 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhA1848M__ExZnUXv8S)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/install_dev_deps.sh#L11)):</sup>
I'll have a think about it and do a followup pull request. I think this is ok for now. But agree, ideal is standard unix flow of error on error, and don't say stuff that doesn't matter. 

---

<sup>**[install_dev_deps.sh, line 20 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAAcwIqqyEYGoElv66)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/install_dev_deps.sh#L20)):</sup>
good to know, thanks! but not needed as this is done within the embedded shell; the script doesn't end up changing the callers working directory. 

---

<sup>**[install_dev_deps.sh, line 30 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAAxnLA8dE-GqSHNjv)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/install_dev_deps.sh#L30)):</sup>
ACK. Will followup later with cleanup/improvements to setup bash script. 

---

<sup>**[README.md, line 21 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAd1OcGqu40TVzmXtF)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/README.md#L21)):</sup>
true. :) done.

---

<sup>**[README.md, line 34 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAd5C5bUmAwp7tsGUy)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/README.md#L34)):</sup>
That's now in build/third_party, so no need. 

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 8 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvSq2dNZ6zbu_62aOu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L8)):</sup>
OK, thinking more about this, the point of the error one is that you are required to spy to complete the mock. So it's really an abstract mock.
the null returner is a SkeletonMock. I think we really just want SkeletonMock as I think about it more. 

---

<sup>**[src/freedom/mocks/jasmine-mock-freedom-module-env.ts, line 10 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAhZOojBhcIDaCINKf)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/src/freedom/mocks/jasmine-mock-freedom-module-env.ts#L10)):</sup>
removed abstract. I didn't think it was very helpful in the end. 

---

<sup>**[src/freedom/typings/freedom-common.d.ts, line 26 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgrdPbDk_X-DdgRht94)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/typings/freedom-common.d.ts#L26)):</sup>
I'm going to followup up separately on a further cleanup of our typing of freedom. There is no name I know of for this object right now. I've added a TODO and issue for it. 

---

<sup>**[src/logging/logging.ts, line 5 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhAPg4zC5buTymFX8Y1)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/src/logging/logging.ts#L5)):</sup>
done

---

<sup>**[src/webrtc/webrtc.ts, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvWFqxQexijn6LrI0H)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/webrtc/webrtc.ts#L1)):</sup>
Pushing the types into peerconnection and datachannel is what I think you're suggesting, but that requires multiple import statements as well, so I don't think it buys anything. 

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)







---

<sup>**[.travis.yml, line 7 \[r8\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhA01Gob0IToatF-kks)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0f6d1e7205480869f1e566aa3a518b08cb622e23/.travis.yml#L7)):</sup>
done.

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)



---

A hearty :+1: once the couple of things we discussed in person are addressed.

Many thanks again.

---

<sup>**[README.md, line 32 \[r10\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhL5CgI_pbKzSH-xZRA)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/8d1f026dba2e6bb12933bc904c41d0bc66d7c470/README.md#L32)):</sup>
`./setup.sh` :-)

---

<sup>**[src/freedom/typings/freedom-common.d.ts, line 132 \[r4\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JgvUi4Kf1gRlM1ot48d)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9763404060d5f1e91ff3456dc631d97e0d1048a9/src/freedom/typings/freedom-common.d.ts#L132)):</sup>
OK, I think I get it now. Many thanks for the helpful comment :-)

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)



---

OK, I think you're going yo like this... (removed webrtc.* files, merging them into datachannel and peerconnection)

---

<sup>**[README.md, line 32 \[r10\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100#-JhL5CgI_pbKzSH-xZRA)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/8d1f026dba2e6bb12933bc904c41d0bc66d7c470/README.md#L32)):</sup>
done.

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)



---

I much prefer this :-)

Any reason for `PeerConnections.PeerConnection` over `PeerConnection.Peerconnection`, though? That seems weird.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)



---

You can't clobber `PeerConnection` at the top level. So if you want to call `PeerConnection` just `PeerConnection` you have to bind the require to something that is not `PeerConnection`. Otherwise you have to refer to it in the file everywhere as `PeerConnection.PeerConnection` which looks even worse to me and isn't isolated to the top of the file. 

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/100)



---

Actually, given a choice between the two, I would choose `PeerConnection.PeerConnection`.

But I don't feel very strongly about it. Still :+1:!

---

PR #101, opened 2015-01-07
Reorder event declarations as recommended by Typescript.

bemasc:
See https://github.com/uProxy/uproxy/issues/766

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/101)


iislucas:
:+1:

PR #107, opened 2015-01-27
emit OFFER and ANSWER signals before others

trevj:
This makes peerconnection always emit an OFFER and ANSWER signal first; more specifically, other signals block until an OFFER or ANSWER signal is sent.

The ordering was easy but I took the opportunity to (finally) add a unit test for `peerconnection`. This required some changes to the API, so I also took the opportunity to introduce some static constructors (`datachannel` already has this, and in general I've found static constructors helpful).

Addresses this issue:
https://github.com/uProxy/uproxy/issues/784

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/107)


trevj:
One thing I'll note is that while the new `emitSignal_` function is not incredibly elegant, I wasn't able to think of a nicer, simpler, way to achieve this. The handler queue interface isn't really - IMO - suited for inheritance and I kinda see us moving towards batching messages inside of one giant message anyway, so this seemed like a reasonable compromise.

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107)

Reviewed files:
- Gruntfile.coffee @ r1
- package.json @ r1
- src/samples/copypaste-freedom-chat/freedom-module.ts @ r1
- src/samples/simple-freedom-chat/freedom-module.ts @ r1
- src/webrtc/datachannel.d.ts @ r1
- src/webrtc/mocks.js @ r1
- src/webrtc/peerconnection.d.ts @ r1
- src/webrtc/peerconnection.spec.ts @ r1
- src/webrtc/peerconnection.ts @ r1

---

Nice fix. Great to see `initiateConnection` removed from config, the factory construction functions, and the small other cleanups. 

Would be good for the test to include a case when the underlying freedom webrtc provider sends the OFFER and CANDIDATES out of order. 

I think there might be a nicer way to write the mock object that distinguishes between spy objects and the mock webrtc provider. Also one small nit on an any type that I think can be more informative. I've provided a link to what I think might be a better style. Lets discuss tomorrow :) 

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 10 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-Jghg9kA96BSrHujfSYx)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L10)):</sup>
I experimented a bit with mock structures and ended up preferring creating a separate file for the mock (because it's often something wanting to be re-used). Using a real class, and having errors returned, or null returned, and then using Jasmine to spy on the objects/functions ended up being easier to understand (and avoid mixing up the spy with the object). Here's an example I did for logging/freedom: 
https://github.com/uProxy/uproxy-lib/blob/iislucas/src/logging/logging.spec.ts
Lets discuss and see what you think. 

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 20 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-JghfNf-vIaCZbzCq9au)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L20)):</sup>
Avoid any types. If you specify the type, it's easier for reader's to understand. This is type should be `() => void`

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 25 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-Jghf_BF_dyYywGtbyYP)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L25)):</sup>
fix any type.

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-Jghh37mdeOxZ1DvfOpH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L46)):</sup>
rather than `{hello: 'world'}` maybe `{ mockLocalDescription: 'blah' }`

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 51 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-JghhN-Xh2AnBScOVGaG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L51)):</sup>
I think this may be better written in a sequential rather than nested style...
```
pc.signalForPeerQueue.setSyncNextHandler((signal:WebRtc.SignallingMessage) => {
      expect(signal.type).toEqual(WebRtc.SignalType.OFFER);
}).then(() => {
  return pc.signalForPeerQueue.setSyncNextHandler((signal:WebRtc.SignallingMessage) => {
      expect(signal.type).toEqual(WebRtc.SignalType.CANDIDATE);
  }); 
}).then(done);
```
Only so that it hints at what to do when you want to do more things in sequence (and avoid making the nesting get worse and worse).

---

<sup>**[src/webrtc/peerconnection.ts, line 156 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-JghenxzgFSgu0eDrcz5)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.ts#L156)):</sup>
Easier to debug if you keep the old naming scheme (appending a random value, or an auto-incremented one).

---

bemasc:
I don't understand this change, or the corresponding bug.  WebRTC will never produce these messages out of order.  If they're getting reordered, it's a bug in our code somewhere, and we should fix it, not re-reorder the messages to put them back in order.

The case where we might want to buffer messages that way is when _receiving_ messages from the remote side, through a transport that is not order-preserving.

iislucas:
What? If WebRTC guarantees order then I agree we should find where it is
going wrong and fix lower down. I am not averse to having a temporary fix
(this code) until then. But we shouldn't only wrap up misbehavior, of
course! :)

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107)







---

<sup>**[src/webrtc/peerconnection.d.ts, line 11 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-JghxeaK3pc17Kx1xvTD)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.d.ts#L11)):</sup>
Is there anything left that uses PeerConnectionConfig, now that you've removed it from the constructor?

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 32 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-JghyVB9ztoQolQaOkTu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L32)):</sup>
I don't think this can happen.  Can freedom really invoke a callback synchronously in the call to on()?  I assume it has to go back to the mainloop, which you could simulate with a zero-delay `setTimeout`.

---

trevj:
Ignoring all other feedback for the moment...great spot, @bemasc! Damn.

I never noticed this before but the messages *still* come out in the wrong order! It turns out that between the time we receive an offer and emit that message, `src/webrtc/peerconnection.ts` has:
 1. called `setDescription`
 1. called `getDescription`

That's here:
https://github.com/uProxy/uproxy-lib/blob/dev/src/webrtc/peerconnection.ts#L370

During that time, an `onicecandidate` is received...and sent. It's a very easy fix but I wanna quickly check with the freedom guys what guarantees we have on receipt of events and promise resolution.

As for the rest of this review...hey, now we have some more unit tests.

trevj:
Correction: this pull request *does* fix the issue...but it is not the most..."foundational" fix.

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107)







---

<sup>**[src/webrtc/peerconnection.spec.ts, line 10 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-Jghg9kA96BSrHujfSYx)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L10)):</sup>
That looks much saner, I'd be extremely happy to use it once your pull request is merged!

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 20 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-JghfNf-vIaCZbzCq9au)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L20)):</sup>
Done.

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 25 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-Jghf_BF_dyYywGtbyYP)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L25)):</sup>
Done.

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 32 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-JghyVB9ztoQolQaOkTu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L32)):</sup>
Not an issue now, I think, but good point (and thanks).

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-Jghh37mdeOxZ1DvfOpH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L46)):</sup>
Removed.

---

<sup>**[src/webrtc/peerconnection.spec.ts, line 51 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-JghhN-Xh2AnBScOVGaG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.spec.ts#L51)):</sup>
Good suggestion but I'm just testing one thing now, so...

---

<sup>**[src/webrtc/peerconnection.ts, line 156 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-JghenxzgFSgu0eDrcz5)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.ts#L156)):</sup>
So, I changed this avoid having to mock out `Crypto` in all the tests. An auto-incremented index is a good idea. Done. Thanks!

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107)







---

<sup>**[src/webrtc/peerconnection.d.ts, line 11 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107#-JghxeaK3pc17Kx1xvTD)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/80b592c55ea1e170e2d00ebf1f2063aa261d9d9a/src/webrtc/peerconnection.d.ts#L11)):</sup>
Done.

---

trevj:
@bemasc, @iislucas : care to take another look?

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/107)

Reviewed files:
- package.json @ r4
- src/webrtc/peerconnection.d.ts @ r3
- src/webrtc/peerconnection.spec.ts @ r2
- src/webrtc/peerconnection.ts @ r3

---

LGTM

---

PR #110, opened 2015-01-29
Edit Jasmine grunt rule to include code coverage.

lucyhe:
Edited Rule.jasmineSpec so that it measures code coverage by default using grunt-template-jasmine-istanbul.

The updated rule maintains the same directory layout assumption as before*, but now users can also pass in a list of source files instead if they'd like.

*e.g. the dir layout assumption for Rule.jasmineSpec 'example_module' would be:
- code to be tested is all of build/example_module/\*\*/\*.js, excluding build/example_module/\*\*/\*spec.js
- tests are build/example_module/\*\*/*spec.js
Now the code coverage would be summarized at build/coverage/example_module/index.html
Allowing users to specify source files will let them customize the files needed to run tests, e.g. mocks needed for testing.

ran grunt test

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/110)

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -86,21 +86,17 @@ module.exports = (grunt) ->
       handler: Rule.jasmineSpec 'handler'
       taskmanager: Rule.jasmineSpec 'taskmanager'
       arraybuffers: Rule.jasmineSpec 'arraybuffers'
-      loggingProvider:
-        src: [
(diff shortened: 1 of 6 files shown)


bemasc:
You'll also need to bump the package version, so uproxy-networking and uproxy can depend on this change.

iislucas:
This was merged into https://github.com/uProxy/uproxy-lib/pull/118
Closing.

PR #116, opened 2015-02-05
set binaryType of new data channels to arraybuffer

trevj:
Along with the various recent changes in freedomjs, this makes proxying work in Firefox.

I'm not sure how best to make this available; do we have to wait until uproxy-networking has been updated to use `require`?

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/116)


iislucas:
Two options: wait until uproxy-networking and uproxy move to require, or back port this to master release to make version 19.01.

The big require pull request didn't update the package version number. Generally, I think we should update package numbers when we cut a new release (buildcop) weekly, or triweekly, and when we do that we review pull requests to decide on version bump size. Require is definitely major (so v20), but if you want to push this to earlier 19.01, then I am ok with that too. That would unblock firefox for uproxy for right now, which would be cool. But if no one is blocked, then it doesn't matter much. 

Make sense?

trevj:
The only way I can think of to "backport" this to a 19.0.1 NPM is by:
 1. checking out an old commit
 1. patching this in locally
 1. rebuilding

Is that what you meant?

willscott:
Sounds right. You can push your back port to a new branch, but i think its
inherently a bit ugly.

--will

trevj:
It's not something I'd like to do very often...since it's blocking Firefox-ification of uproxy-networking, I think I'll give it a whirl. Thanks.

PR #120, opened 2015-02-12
Mark some TCP info attributes as optional

bemasc:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/120)


trevj:
:+1:

PR #125, opened 2015-02-20
Make the logging interface slightly more usable

jpevarnek:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/125)

### DIFF
--- a/src/logging/logging.spec.ts
+++ b/src/logging/logging.spec.ts
@@ -95,7 +95,7 @@ describe('Client logging shim using Freedom', () => {
       });
     });
 
-    it('stringify objections', (done) => {
+    it('stringify objects', (done) => {
       var log = new Logging.Log('tag');
       var obj = { foo: 'bar' };
 
--- a/src/logging/logging.spec.ts
+++ b/src/logging/logging.spec.ts
@@ -2,52 +2,89 @@
 /// <reference path='../../build/third_party/freedom-typings/freedom-module-env.d.ts' />
 
 import freedomMocker = require('../freedom/mocks/mock-freedom-in-module-env');
+import freedomTypes = require('freedom.types');
 import Logging = require('./logging');
 
-describe("Client logging shim using Freedom", () => {
-
+describe('Client logging shim using Freedom', () => {
   beforeEach(() => {
     // Reset the mock freedom environment.
     freedom = freedomMocker.makeMockFreedomInModuleEnv();
   });
 
-  it('A new Logging.Log forwards all logging to the named freedom core logger.',
-      (done) => {
-    var mockLoggerPromise = Promise.resolve(jasmine.createSpyObj(
-      'tag1', ['debug', 'log', 'info', 'warn', 'error']));
-    var getLoggerSpy = spyOn(freedom.core(), "getLogger");
-    getLoggerSpy.and.returnValue(mockLoggerPromise);
+  describe('tag tests', () => {
+    it('Get logger with tag tag1', (done) => {
+      var mockLoggerPromise = Promise.resolve(jasmine.createSpyObj(
+        'tag1', ['debug', 'log', 'info', 'warn', 'error']));
+      var getLoggerSpy = spyOn(freedom.core(), "getLogger");
+      getLoggerSpy.and.returnValue(mockLoggerPromise);
 
-    var log1 = new Logging.Log('tag1');
-    expect(getLoggerSpy).toHaveBeenCalledWith('tag1');
+      var log1 = new Logging.Log('tag1');
+      expect(getLoggerSpy).toHaveBeenCalledWith('tag1');
+      done();
+    });
 
-    log1.error('test-error-string');
-    log1.debug('test-debug-string');
+    it('Get logger with tag tag2', (done) => {
+      var mockLoggerPromise = Promise.resolve(jasmine.createSpyObj(
+        'tag1', ['debug', 'log', 'info', 'warn', 'error']));
+      var getLoggerSpy = spyOn(freedom.core(), "getLogger");
+      getLoggerSpy.and.returnValue(mockLoggerPromise);
 
-    mockLoggerPromise.then((mockLogger) => {
-      expect(mockLogger.debug).toHaveBeenCalledWith('test-debug-string');
-      expect(mockLogger.error).toHaveBeenCalledWith('test-error-string');
-      expect(mockLogger.log).not.toHaveBeenCalledWith('test-error-string');
+      var log2 = new Logging.Log('tag2');
+      expect(getLoggerSpy).not.toHaveBeenCalledWith('tag1');
+      expect(getLoggerSpy).toHaveBeenCalledWith('tag2');
       done();
     });
   });
 
-  it('Collapses arguments into flattened messages', (done) => {
-    var mockLoggerPromise = Promise.resolve(jasmine.createSpyObj(
-      'tag1', ['debug', 'log', 'info', 'warn', 'error']));
-    var getLoggerSpy = spyOn(freedom.core(), "getLogger");
-    getLoggerSpy.and.returnValue(mockLoggerPromise);
+  describe('Log messages', () => {
+    var mockLoggerPromise :Promise<freedomTypes.Logger>;
 
-    var log2 = new Logging.Log('tag2');
-    log2.info('%1 pinged %2 with id=%3', ['Bob', 'Alice', '123456']);
-    expect(getLoggerSpy).not.toHaveBeenCalledWith('tag1');
-    expect(getLoggerSpy).toHaveBeenCalledWith('tag2');
+    beforeEach(() => {
+      var mockLogger = jasmine.createSpyObj<freedomTypes.Logger>('tag',
+          ['debug', 'log', 'info', 'warn', 'error']);
+      mockLoggerPromise = Promise.resolve(mockLogger);
 
-    mockLoggerPromise.then((mockLogger) => {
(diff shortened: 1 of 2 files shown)


src/logging/logging.ts
```diff
@@ -5,37 +5,29 @@ import freedomTypes = require('freedom.types');
...
-  var formatted_msg = msg;
-  if (args && args.length) {
-    for (var i = 0; i < args.length; i++) {
-      formatted_msg = formatted_msg.replace('%' + (i + 1), args[i]);
-    }
-  }
-  return formatted_msg;
-}
+  var msg = '';
+  var arg :any;
```
trevj: You can declare this closer to where you use it, i.e. in the loop body.

src/logging/logging.ts
```diff
@@ -5,37 +5,29 @@ import freedomTypes = require('freedom.types');
...
 
-interface loggable {
-  (logger: freedomTypes.Logger): void;
-}
+  for (var i = 0; i < args.length; i++) {
+    arg = args[i];
+    if ('string' !== typeof(arg)) {
+      arg = JSON.stringify(arg);
```
trevj: What happens if I supply just a primitive?

src/logging/logging.ts
```diff
@@ -5,37 +5,29 @@ import freedomTypes = require('freedom.types');
...
-}
+  var msg = '';
+  var arg :any;
 
-interface loggable {
-  (logger: freedomTypes.Logger): void;
-}
+  for (var i = 0; i < args.length; i++) {
+    arg = args[i];
+    if ('string' !== typeof(arg)) {
```
trevj: I think you also need to consider the case where `arg` has been created via `new String`.

src/logging/logging.ts
```diff
@@ -43,20 +35,45 @@ export class Log {
   constructor(private tag_:string) {
     this.logger = freedom.core().getLogger(this.tag_);
   }
+
+  private log_ = (level :string, args :any[]) :void => {
+    var message :string;
+
+    if (2 === args.length && 'string' === typeof(args[0]) && Array.isArray(args[1])) {
```
trevj: Again, I think you need to consider strings created via `new String()`.

trevj: FYI, we use the form 'x == 1' much more than '1 == x' in uProxy code (personally I find it much much easier to read).

src/logging/logging.ts
```diff
@@ -43,20 +35,45 @@ export class Log {
...
+      args = [args[0]].concat(args[1].slice());
```
trevj: FYI, you couls use `unshift` here.

src/logging/logging.ts
```diff
@@ -43,20 +35,45 @@ export class Log {
   constructor(private tag_:string) {
     this.logger = freedom.core().getLogger(this.tag_);
   }
+
+  private log_ = (level :string, args :any[]) :void => {
+    var message :string;
```
trevj: It's often a good idea to declare variables closer to where they're actually used. In this case you can save a few lines and have the compiler infer the type for you.

src/logging/logging.spec.ts
```diff
@@ -2,52 +2,111 @@
...
+      var log = new Logging.Log('tag');
+
+      log.info('%1', 'foo', 'bar');
+      mockLoggerPromise.then((mockLogger :freedomTypes.Logger) => {
+        expect(mockLogger.info).toHaveBeenCalledWith('foo bar');
+        done();
+      });
+    });
+
+    it('stringify objections', (done) => {
```
trevj: `objects`? :-)

src/logging/logging.ts
```diff
@@ -5,37 +5,29 @@ import freedomTypes = require('freedom.types');
...
-    } else if (level === 'info') {
-      logger.info(message);
-    } else if (level === 'warn') {
-      logger.warn(message);
-    } else if (level === 'error') {
-      logger.error(message);
+    if (-1 !== msg.indexOf('%' + i)) {
+      msg = msg.replace('%' + i, arg);
+    } else {
+      if (msg.length) {
```
trevj: Better/clearer/safer to use `msg.length > 0`, I think.

src/logging/logging.ts
```diff
@@ -5,37 +5,29 @@ import freedomTypes = require('freedom.types');
...
-  return (logger: freedomTypes.Logger) => {
-    if (level === 'debug') {
-      logger.debug(message);
-    } else if (level === 'info') {
-      logger.info(message);
-    } else if (level === 'warn') {
-      logger.warn(message);
-    } else if (level === 'error') {
-      logger.error(message);
+    if (-1 !== msg.indexOf('%' + i)) {
```
trevj: Again, we generally use the reverse check in uProxy code.

trevj:
Nice. I have often forgotten to supply an array, to much confusion. A few things to look at before it's ready.

src/logging/logging.ts
jpevarnek: I completely agree and normally do it that way, I was trying to follow https://docs.google.com/a/google.com/document/d/1ZV70oTNWO00bs0RcMwqr3FvKJXA6smYQ2kZ8Zr8YZnw/edit#heading=h.a3wjgdv24nv9, is that no longer accurate?

src/logging/logging.ts
jpevarnek: Okay, I've been used to putting variable declarations at the top of functions because it seems to make javascript scoping more clear but I'm fine with either.

src/logging/logging.ts
trevj: Huh. I wasn't aware of that.

I guess this is fine for now but I just started a thread on `uproxy-eng`.

src/logging/logging.ts
jpevarnek: `JSON.stringify` acts pretty much as `::toString` in that case.

src/logging/logging.ts
jpevarnek: Done (and below)

src/logging/logging.ts
trevj: Oh! Cool.

src/logging/logging.ts
jpevarnek: I might be misunderstanding this a bit, wouldn't that need to be:
```
var foo = args[1].slice();
foo.unshift(args[0]);
args = foo;
```

src/logging/logging.spec.ts
jpevarnek: Done.

src/logging/logging.ts
jpevarnek: Done.

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/125)

Reviewed files:
- src/logging/logging.spec.ts @ r2
- src/logging/logging.ts @ r2

---

Nice to see energy on making logging better! 

The use of switch and also the ability to use nary arguments instead of lists is good. 

Probably want a try/catch around JSON.stringify. 

Also: lets force a log statement to have at least one argument according to its type. 

Don't use `any` type when you can use `Object` (for logging arguments, we only require them to be Object, and I think that's better, it's equivalent, I believe to the `{}` type.).

---

<sup>**[src/logging/logging.ts, line 63 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/125#-Jiz7PqjgKMAqraOXlBb)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/60a4d5c8bcdec573821859fd9ac87a627483ddee/src/logging/logging.ts#L63)):</sup>
Not sure we want to allow a debug call with no arguments.... ? 
Maybe  a better type is: `(msg: string, ...args:any[]) :void` ?

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/125)







---

<sup>**[src/logging/logging.ts, line 63 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/125#-Jiz7PqjgKMAqraOXlBb)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/60a4d5c8bcdec573821859fd9ac87a627483ddee/src/logging/logging.ts#L63)):</sup>
Actually, `(msg: Object, ...args:Object[]) :void` is better. :) 

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/125)



---

:+1:

---

<sup>**[src/logging/logging.ts, line 28 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/125#-Jj6Jg6xZIwTxXkvB10B)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/f61a896730e6931320b44a16b834294d50ec2327/src/logging/logging.ts#L28)):</sup>
nit: i think the cast has no effect here

---

jpevarnek:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/125)







---

<sup>**[src/logging/logging.ts, line 28 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/125#-Jj6Jg6xZIwTxXkvB10B)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/f61a896730e6931320b44a16b834294d50ec2327/src/logging/logging.ts#L28)):</sup>
Correct, it was needed to silence Typescript though.

---

PR #126, opened 2015-02-20
Check the buffered amount before sending each queued block.

bemasc:
Tested using (forthcoming) load tests in uproxy-networking.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/126)

### DIFF
--- a/src/webrtc/datachannel.ts
+++ b/src/webrtc/datachannel.ts
@@ -127,7 +127,7 @@ export class DataChannelClass implements DataChannel {
     });
     this.onceOpened.then(() => {
       this.opennedSuccessfully_ = true;
-      this.toPeerDataQueue_.setHandler(this.handleSendDataToPeer_);
+      this.toPeerDataQueue_.setNextHandler(this.handleSendDataToPeer_);
     });
     this.onceClosed.then(() => {
         if(!this.opennedSuccessfully_) {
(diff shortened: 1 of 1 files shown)


iislucas:
:+1: 

Would love to have this land, with updated package.json, so I can include it in the release I plan to cut later today :)

src/webrtc/datachannel.ts
```diff
@@ -249,7 +249,14 @@ export class DataChannelClass implements DataChannel {
         setTimeout(this.conjestionControlSendHandler, 20);
       } else {
         if(!this.toPeerDataQueue_.isHandling()) {
-          this.toPeerDataQueue_.setHandler(this.handleSendDataToPeer_);
+          // This processes one block from the queue, which (in Chrome) is
```
trevj: If I'm reading this correctly, you can remove the `if` check because the queue will never have a handler. Is that correct?

trevj:
So, we can punt on removing the `if` block. Once you've tested this change I think it's good to submit. :+1: 

Then, me or @iislucas will get this into a 19.x.x NPM.

Thanks!

src/webrtc/datachannel.ts
bemasc: Done.

PR #127, opened 2015-02-24
Code cleanup; simpler Gruntfile; freedom type improvement

iislucas:
* Code cleanup of some headers (needed for forthcoming uproxy-networking cleanup)
* Simplified Gruntfile (following style of forthcoming uproxy-networking cleanup)
* Freedom type improvement (needed for forthcoming uproxy-networking cleanup)
* Package bump to remind us that this is not the v19 branch
* Removes magic string-literals in logging provider (use constants instead)

TESTED: 
 * grunt 
 * Ran & tested the sample apps

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/127)


trevj:
Seems like a big improvement.

:+1:

PR #128, opened 2015-02-25
Updates in preparation for uproxy-networking requireify

iislucas:
Based on this doc: https://docs.google.com/document/d/1H3ac_gZFJfhUGyumCbAKhXSa-_UsusDmsAUqXe4h-JQ/edit#

 * Returned to sym-link before complication (long story...)
 * Generalized library copying function in common-grunt-tools
 * Added third_party build directory as a parameter to common-grunt-tools
 * Updated readme and updated paths

TESTED: 
 * grunt
 * ran sample apps

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/128)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,4 @@
 build/*
 node_modules/*
 npm-debug.log
-src/.baseDir.ts
 tmp/*
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -69,9 +69,11 @@ path = require 'path'
 #-------------------------------------------------------------------------
 devBuildDir = 'build/dev/uproxy-lib'
 thirdPartyBuildDir = 'build/third_party'
+localLibsDestPath = 'uproxy-lib'
 Rule = new rules.Rule({
   devBuildDir: devBuildDir,
-  thirdPartyBuildDir: thirdPartyBuildDir
+  thirdPartyBuildDir: thirdPartyBuildDir,
+  localLibsDestPath: localLibsDestPath
 });
 
 module.exports = (grunt) ->
@@ -135,10 +137,10 @@ module.exports = (grunt) ->
       # Rule.copyLibs [npmModules], [localDirectories], [thirdPartyDirectories]
       libsForSimpleFreedomChat:
         Rule.copyLibs ['freedom'], ['loggingprovider'], [],
-          'samples/simple-freedom-chat/lib'
+          'samples/simple-freedom-chat/'
       libsForCopypasteFreedomChat:
         Rule.copyLibs ['freedom'], ['loggingprovider'], [],
-          'samples/copypaste-freedom-chat/lib'
+          'samples/copypaste-freedom-chat/'
 
     # Typescript rules
     ts:
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -6,7 +6,7 @@ TaskManager = require './build/tools/taskmanager'
 taskManager = new TaskManager.Manager();
 
 # Makes the base development build, excludes sample apps.
-taskManager.add 'base-dev', [
+taskManager.add 'base', [
   'symlink:typescriptSrc'
   'copy:third_party'
   'copy:dev'
@@ -15,22 +15,23 @@ taskManager.add 'base-dev', [
   'browserify:loggingProvider'
 ]
 
-# Makes the development build, includes sample apps.
-taskManager.add 'dev', [
-  'base-dev'
+# Makes all sample apps.
+taskManager.add 'samples', [
+  'base'
   'simpleFreedomChat'
   'copypasteFreedomChat'
 ]
 
 # Makes the distribution build.
 taskManager.add 'dist', [
-  'dev',
+  'base',
+  'samples',
   'copy:dist'
 ]
 
 # Build the simple freedom chat sample app.
 taskManager.add 'simpleFreedomChat', [
-  'base-dev'
+  'base'
   'copy:libsForSimpleFreedomChat'
   'browserify:simpleFreedomChatMain'
   'browserify:simpleFreedomChatFreedomModule'
@@ -38,7 +39,7 @@ taskManager.add 'simpleFreedomChat', [
 
 # Build the copy/paste freedom chat sample app.
 taskManager.add 'copypasteFreedomChat', [
-  'base-dev'
+  'base'
   'copy:libsForCopypasteFreedomChat'
   'browserify:copypasteFreedomChatMain'
   'browserify:copypasteFreedomChatFreedomModule'
@@ -46,7 +47,7 @@ taskManager.add 'copypasteFreedomChat', [
 
 # Run unit tests
 taskManager.add 'unit_tests', [
-  'base-dev'
+  'base'
   'browserify:arraybuffersSpec'
   'browserify:handlerSpec'
   'browserify:buildToolsTaskmanagerSpec'
@@ -60,19 +61,19 @@ taskManager.add 'unit_tests', [
 taskManager.add 'test', ['unit_tests']
 
 # Default task, build dev, run tests, make the distribution build.
-taskManager.add 'default', ['dev', 'unit_tests', 'dist']
+taskManager.add 'default', ['base', 'unit_tests', 'dist']
 
 #-------------------------------------------------------------------------
 rules = require './build/tools/common-grunt-rules'
 path = require 'path'
 
 #-------------------------------------------------------------------------
-devBuildDir = 'build/dev/uproxy-lib'
-thirdPartyBuildDir = 'build/third_party'
+devBuildPath = 'build/dev/uproxy-lib'
+thirdPartyBuildPath = 'build/third_party'
 localLibsDestPath = 'uproxy-lib'
 Rule = new rules.Rule({
-  devBuildDir: devBuildDir,
-  thirdPartyBuildDir: thirdPartyBuildDir,
+  devBuildPath: devBuildPath,
+  thirdPartyBuildPath: thirdPartyBuildPath,
   localLibsDestPath: localLibsDestPath
 });
 
@@ -81,15 +82,18 @@ module.exports = (grunt) ->
     pkg: grunt.file.readJSON 'package.json'
 
     symlink:
+      # Sym-link typeScript sources from src/ to build/dev/uproxy-lib/
       typescriptSrc:
         files: [{
           expand: true
           overwrite: false
           cwd: 'src'
           src: ['**/*.ts']
-          dest: devBuildDir
+          dest: devBuildPath
         }]
 
+    # Sym-links are prefered, but chrome apps require real files. So we copy all
+    # non-compiled files.
     copy:
       # Copy releveant non-typescript files to dev build.
       dev:
@@ -99,13 +103,13 @@ module.exports = (grunt) ->
               expand: true,
               cwd: 'src/',
               src: ['**/*', '!**/*.ts'],
-              dest: devBuildDir,
+              dest: devBuildPath,
               onlyIf: 'modified'
           }
         ]
-      # Copy |third_party| to dev: this is so that there is a common
-      # |thirdPartyBuildDir| location to reference typescript
-      # definitions for ambient contexts.
+      # Copy |third_party| to build folder: this is so that there is a common
+      # |thirdPartyBuildPath| location to reference typescript definitions for
+      # ambient contexts.
       third_party:
         files: [
           {
@@ -113,7 +117,7 @@ module.exports = (grunt) ->
               expand: true,
               cwd: 'third_party'
               src: ['**/*'],
-              dest: thirdPartyBuildDir,
+              dest: thirdPartyBuildPath,
               onlyIf: 'modified'
           }
         ]
@@ -123,7 +127,7 @@ module.exports = (grunt) ->
           {
               nonull: true,
               expand: true,
-              cwd: devBuildDir,
+              cwd: devBuildPath,
               src: ['**/*',
                     '!**/*.spec.js',
                     '!**/*.ts',
@@ -148,12 +152,10 @@ module.exports = (grunt) ->
       # build directory.
       devInModuleEnv:
         src: [
-          devBuildDir + '/**/*.ts'
-          '!' + devBuildDir + '/**/*.core-env.ts'
-          '!' + devBuildDir + '/**/*.core-env.spec.ts'
+          devBuildPath + '/**/*.ts'
+          '!' + devBuildPath + '/**/*.core-env.ts'
+          '!' + devBuildPath + '/**/*.core-env.spec.ts'
         ]
-        #outDir: 'build/dev/'
-        #baseDir: 'src'
         options:
           target: 'es5'
           comments: true
@@ -166,11 +168,9 @@ module.exports = (grunt) ->
       # build directory.
       devInCoreEnv:
         src: [
-          devBuildDir + '/**/*.core-env.ts'
-          devBuildDir + '/**/*.core-env.spec.ts'
+          devBuildPath + '/**/*.core-env.ts'
+          devBuildPath + '/**/*.core-env.spec.ts'
         ]
-        #outDir: 'build/dev/'
-        #baseDir: 'src'
         options:
           target: 'es5'
           comments: true
@@ -207,7 +207,6 @@ module.exports = (grunt) ->
     clean:
       build:
         [ 'build/dev', 'build/dist'
-          # Note: '.tscache/' is created by grunt-ts.
           '.tscache/']
 
   #-------------------------------------------------------------------------
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -7,6 +7,7 @@ taskManager = new TaskManager.Manager();
 
 # Makes the base development build, excludes sample apps.
 taskManager.add 'base-dev', [
+  'symlink:typescriptSrc'
   'copy:third_party'
   'copy:dev'
   'ts:devInModuleEnv'
@@ -30,7 +31,7 @@ taskManager.add 'dist', [
 # Build the simple freedom chat sample app.
 taskManager.add 'simpleFreedomChat', [
   'base-dev'
-  'copy:freedomLibsForSimpleFreedomChat'
+  'copy:libsForSimpleFreedomChat'
   'browserify:simpleFreedomChatMain'
   'browserify:simpleFreedomChatFreedomModule'
 ]
@@ -38,7 +39,7 @@ taskManager.add 'simpleFreedomChat', [
 # Build the copy/paste freedom chat sample app.
 taskManager.add 'copypasteFreedomChat', [
   'base-dev'
-  'copy:freedomLibsForCopypasteFreedomChat'
+  'copy:libsForCopypasteFreedomChat'
   'browserify:copypasteFreedomChatMain'
   'browserify:copypasteFreedomChatFreedomModule'
 ]
@@ -66,12 +67,26 @@ rules = require './build/tools/common-grunt-rules'
 path = require 'path'
 
 #-------------------------------------------------------------------------
-devBuildDir = 'build/dev'
-Rule = new rules.Rule({devBuildDir: devBuildDir});
+devBuildDir = 'build/dev/uproxy-lib'
+thirdPartyBuildDir = 'build/third_party'
+Rule = new rules.Rule({
+  devBuildDir: devBuildDir,
+  thirdPartyBuildDir: thirdPartyBuildDir
+});
 
 module.exports = (grunt) ->
   config =
     pkg: grunt.file.readJSON 'package.json'
+
+    symlink:
+      typescriptSrc:
+        files: [{
+          expand: true
+          overwrite: false
+          cwd: 'src'
+          src: ['**/*.ts']
+          dest: devBuildDir
+        }]
 
     copy:
       # Copy releveant non-typescript files to dev build.
@@ -87,15 +102,16 @@ module.exports = (grunt) ->
           }
         ]
       # Copy |third_party| to dev: this is so that there is a common
-      # |build/third_party| location to reference typescript
+      # |thirdPartyBuildDir| location to reference typescript
       # definitions for ambient contexts.
       third_party:
         files: [
           {
               nonull: true,
               expand: true,
-              src: ['third_party/**/*'],
-              dest: 'build/',
+              cwd: 'third_party'
+              src: ['**/*'],
+              dest: thirdPartyBuildDir,
               onlyIf: 'modified'
           }
         ]
@@ -108,18 +124,20 @@ module.exports = (grunt) ->
               cwd: devBuildDir,
               src: ['**/*',
                     '!**/*.spec.js',
-                    '!**/*.spec.*.js'],
+                    '!**/*.ts',
+                    '**/*.d.ts'],
               dest: 'build/dist/',
               onlyIf: 'modified'
           }
         ]
 
       # Copy the freedom output file to sample apps
-      freedomLibsForSimpleFreedomChat:
-        Rule.copyFreedomLibs 'freedom', ['loggingprovider'],
+      # Rule.copyLibs [npmModules], [localDirectories], [thirdPartyDirectories]
+      libsForSimpleFreedomChat:
+        Rule.copyLibs ['freedom'], ['loggingprovider'], [],
           'samples/simple-freedom-chat'
-      freedomLibsForCopypasteFreedomChat:
-        Rule.copyFreedomLibs 'freedom', ['loggingprovider'],
+      libsForCopypasteFreedomChat:
+        Rule.copyLibs ['freedom'], ['loggingprovider'], [],
           'samples/copypaste-freedom-chat/'
 
     # Typescript rules
@@ -128,12 +146,12 @@ module.exports = (grunt) ->
       # build directory.
       devInModuleEnv:
         src: [
-          'src/**/*.ts'
-          '!src/**/*.core-env.ts'
-          '!src/**/*.core-env.spec.ts'
+          devBuildDir + '/**/*.ts'
+          '!' + devBuildDir + '/**/*.core-env.ts'
+          '!' + devBuildDir + '/**/*.core-env.spec.ts'
         ]
-        outDir: 'build/dev/'
-        baseDir: 'src'
+        #outDir: 'build/dev/'
+        #baseDir: 'src'
         options:
           target: 'es5'
           comments: true
@@ -146,11 +164,11 @@ module.exports = (grunt) ->
       # build directory.
       devInCoreEnv:
         src: [
-          'src/**/*.core-env.ts'
-          'src/**/*.core-env.spec.ts'
+          devBuildDir + '/**/*.core-env.ts'
+          devBuildDir + '/**/*.core-env.spec.ts'
         ]
-        outDir: 'build/dev/'
-        baseDir: 'src'
+        #outDir: 'build/dev/'
+        #baseDir: 'src'
         options:
           target: 'es5'
           comments: true
@@ -187,9 +205,8 @@ module.exports = (grunt) ->
     clean:
       build:
         [ 'build/dev', 'build/dist'
-          # Note: 'src/.baseDir.ts' and '.tscache/' are created by grunt-ts.
-          '.tscache/'
-          'src/.baseDir.ts' ]
+          # Note: '.tscache/' is created by grunt-ts.
+          '.tscache/']
 
   #-------------------------------------------------------------------------
   grunt.initConfig config
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -7,7 +7,6 @@ taskManager = new TaskManager.Manager();
 
 # Makes the base development build, excludes sample apps.
 taskManager.add 'base', [
-  'symlink:typescriptSrc'
   'copy:third_party'
   'copy:dev'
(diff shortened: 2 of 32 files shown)


trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128)



---

I have a bunch of comments but they're mostly concerning comments, variable naming, and some tidying suggestions.

The only fundamental thing I'm concerned with is that it feels like we're still bending over backwards to make sample apps happy. They are very important (and of course uProxy itself is a kind of sample app) but I kinda liked the old approach of just dumping *everything* into a sample app's `lib/ `folder. If we are not going to distribute sample apps in `dist`/ then I wonder if some symlinking magic could save us a bunch of work in that respect? Something to think about.

---

<sup>**[Gruntfile.coffee, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5S_bggNHVr4PRS6uG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L9)):</sup>
What do you think of renaming this `base`?

---

<sup>**[Gruntfile.coffee, line 70 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5YZWcPvsh1YxUQ85N)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L70)):</sup>
We have a weird mix of `BuildDir` and `Path` suffixes here, and in the `RuleConfig` object in `common-grunt-rules.ts`.

Personally, I much prefer Path which give us something like:
```coffeescript
# Destination for compiled TypeScript.
outPath = 'build/dev/uproxy-lib'
# Path to third_party directory.
thirdPartyPath = 'build/third_party'
# (description)
libPath = 'uproxy-lib'
```

---

<sup>**[Gruntfile.coffee, line 83 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5WwH97R6bKPs9R-A5)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L83)):</sup>
So you're symlinking the TypeScript sources. This is fine but why do we symlink sources while copying non-TypeScript files and `third_party`?

---

<sup>**[Gruntfile.coffee, line 83 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5XYZo0lkSuL4EDyhS)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L83)):</sup>
We could do with a comment here.

Generally, the comments for the tasks comprising `base-dev` bother me: would we ever copy non-relevant files? :-) Suggest we make them something like:
```
# Copies non-TypeScript sources from src/ to build/dev/uproxy-lib/.
```

(or use the `$xxxPath` variables)

---

<sup>**[Gruntfile.coffee, line 84 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5SMUJfhyLqLCjENyq)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L84)):</sup>
I think you could rename this `symlink:src` without sacrificing any clarity.

---

<sup>**[Gruntfile.coffee, line 155 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5cqJQZJQrsMngsmkI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L155)):</sup>
remove commented line (and elsewhere in this file)

---

<sup>**[Gruntfile.coffee, line 210 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5Y1fRtbry2aBzUwWV)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L210)):</sup>
Crazy formatting here: move the comment before the array declaration. In fact, I don't think this comment adds much (and why comment just this line? seems like the easiest path of the three to figure out)

---

<sup>**[setup.sh, line 23 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5TXIKgw4lqFLc6-CF)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L23)):</sup>
Would `ln -f` achieve what you want?

---

<sup>**[setup.sh, line 26 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5U75VLAJXYaVE5rdH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L26)):</sup>
Why can't you just perform the linking and compilation in `build/tools`?

Have you considered making `src/build-tools` a top-level folder? That might help clarify that it sits apart from the other source code-containing folders.

---

<sup>**[setup.sh, line 31 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5UX743X5FZd4kjeni)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L31)):</sup>
The `cd`s look like a typo.

Why did you remove the `cd $ROOT_DIR` at line 38? It looks like every function depends on it.

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5_Ub6rZrXPvdaFLUG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L7)):</sup>
Again, weird mix of `buildDir` and `Path` suffixes. Comments needed, too.

```typescript
export interface RuleConfig {
  // Destination for compiled TypeScript.
  outPath :string;
  // Path to third_party directory.
  thirdPartyPath :string;
  // Destination for libraries.
  libPath :string;
}
```

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 110 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5d1DEXQbkmYHxKu0v)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L110)):</sup>
This function is *really* packing in the functionality. That you felt it useful to add a comment to the `Gruntfile` describing the arguments suggests it's overly-overloaded.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128)



---

Thanks for the comments! 

I think the only extra thing for sample apps now is the copyLibs function. I prefer explicit dependencies being expressed; in time I hope to move to gulp and a simpler system which just has to be told the dependencies and not much more. But for now, this is, I think a pragmatic in between. 

PTAL

---

<sup>**[Gruntfile.coffee, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5S_bggNHVr4PRS6uG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L9)):</sup>
good idea, in base I'll rename dev to `samples` and this to `base`

---

<sup>**[Gruntfile.coffee, line 70 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5YZWcPvsh1YxUQ85N)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L70)):</sup>
thanks! renamed to end in Path.

---

<sup>**[Gruntfile.coffee, line 83 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5WwH97R6bKPs9R-A5)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L83)):</sup>
I'd prefer sym-links for all. But chrome apps don't allow sym-links. I guess I could sym-link everything and write special case for copying for samples... ? Or maybe sym-link for all, and put samples in dist and copy stuff there so it is only expected to work in dist? 

Sym-links are not always handled as well as real files (e.g. in Sublime Text), so I think I'm ok with the funny compromise we have with sym-links for compiled stuff, and copies for everything else... but I don't feel super strongly. 

---

<sup>**[Gruntfile.coffee, line 83 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5XYZo0lkSuL4EDyhS)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L83)):</sup>
Not sure I understand... but I'm now using xxxPath vars, and also added (a variant of) that comment for the typescriptSrc rule.

---

<sup>**[Gruntfile.coffee, line 84 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5SMUJfhyLqLCjENyq)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L84)):</sup>
There's some JS code and other stuff people consider source code, so I'll stick with being more verbose.

---

<sup>**[Gruntfile.coffee, line 155 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5cqJQZJQrsMngsmkI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L155)):</sup>
Done.

---

<sup>**[Gruntfile.coffee, line 210 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5Y1fRtbry2aBzUwWV)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L210)):</sup>
Done.

---

<sup>**[setup.sh, line 23 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5TXIKgw4lqFLc6-CF)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L23)):</sup>
Done.

---

<sup>**[setup.sh, line 26 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5U75VLAJXYaVE5rdH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L26)):</sup>
Yeah, I've pondered it; can't make up my mind. Until I have a more compelling reason to change I'll stick with it as it. I like that generic tools like task-manager as in a place for generic tools. I like the uniformity of the code layout. But it's not a strong argument. :) 

---

<sup>**[setup.sh, line 31 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5UX743X5FZd4kjeni)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L31)):</sup>
Thanks for type catch! Luckily no important file was named `cd`! (Done)

'cd $ROOT' gets printed out when you ask for help; so the output looks kind of odd. Slightly prefer it this way (I think it's a little more modular for independent calls too). 

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5_Ub6rZrXPvdaFLUG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L7)):</sup>
Done.

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 110 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5d1DEXQbkmYHxKu0v)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L110)):</sup>
Yeah, not something I love. Could use a named object to make arguments more obvious. I still find it saves a lot of gruntfile to have this. Generally, I plan to move to a name-structured build system so that you don't have to write nonsense like this. Just specify files as dependencies and nothing more. But that will have to wait on gulp, so I'm not planning to do a lot more on this right now. 

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128)







---

<sup>**[Gruntfile.coffee, line 71 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5YZWcPvsh1YxUQ85N)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L71)):</sup>
So, I still think we can do a lot better with naming.

For example, what's a "local lib"? What would a "remote lib" be? `devBuildPath` seems okay, if verbose -- does that imply that `thirdPartyBuildPath` is where we build third party code?

---

<sup>**[Gruntfile.coffee, line 84 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5WwH97R6bKPs9R-A5)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L84)):</sup>
The build system might be easier to grok if we just make copies of *everything*. Come to think of it, if we just copied all of `src/` that would allow us merge `symlink:typeScriptSrc` and `copy:dev` into one that just copies the entire `src/` folder. Thoughts?

---

<sup>**[Gruntfile.coffee, line 84 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5XYZo0lkSuL4EDyhS)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L84)):</sup>
I meant the comments at lines 98 and 124, e.g.:
```
 # Copy releveant non-typescript files to dev build.
```

These could use the same commenting love as `typescriptSrc`.

BTW, I've never seen "symlink" hyphenated.

---

<sup>**[Gruntfile.coffee, line 86 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5SMUJfhyLqLCjENyq)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L86)):</sup>
So, see above comment for how we could merge this with `copy:dev`.

If you don't like that idea, I just realised there's also no obvious link between `symlink:typeScriptSrc` and `copy:dev` even though they operate on the same folder...`copy:nonTypeScriptSrc`?

---

<sup>**[Gruntfile.coffee, line 209 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5Y1fRtbry2aBzUwWV)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L209)):</sup>
Sorry to be nitpicky but these could all fit on one line. Or you could follow the array convention set by the rest of the file.

---

<sup>**[setup.sh, line 24 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5TXIKgw4lqFLc6-CF)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L24)):</sup>
And remove the `|| true`, right?

---

<sup>**[setup.sh, line 27 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5U75VLAJXYaVE5rdH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L27)):</sup>
Honestly, I think the messing about here is strong enough argument.

---

<sup>**[setup.sh, line 32 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5UX743X5FZd4kjeni)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L32)):</sup>
But I think you (really really) need to add `runCmd "cd $ROOT_DIR` to `clean()`, no?

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5_Ub6rZrXPvdaFLUG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L7)):</sup>
Comments are really needed here.

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 110 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5d1DEXQbkmYHxKu0v)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L110)):</sup>
I wouldn't hold my breath on moving to another build system, so I really think we need to do better now.

A named argument might make it less error-prone to call but so would some comments explaining the arguments.

I can see how this saves several lines in the `Gruntfile` (presumably many in uproxy-networking) so, at a minimum, let's add comments explaining the arguments.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128)



---

OK, more comments and thinking about names.  PTAL

---

<sup>**[Gruntfile.coffee, line 71 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5YZWcPvsh1YxUQ85N)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L71)):</sup>
ok, lets do it!

Building involves linking code form third parties; that code lives in third party build path. 

For now I've added some comments to help us think of better names. Let me know what you think.

---

<sup>**[Gruntfile.coffee, line 84 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5WwH97R6bKPs9R-A5)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L84)):</sup>
Yeah, I'm ok with that. done.

---

<sup>**[Gruntfile.coffee, line 84 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5XYZo0lkSuL4EDyhS)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L84)):</sup>
Removed symlinking altogether, done now in a single copy. This has the downside that you have to remember to edit your typescript in the src directory, not in the build directory (this is the motivation for symlinking, but I'm also ok with accepting this for the simpler gruntfile).

---

<sup>**[Gruntfile.coffee, line 86 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5SMUJfhyLqLCjENyq)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L86)):</sup>
merged.  It's a little simpler now. 

---

<sup>**[Gruntfile.coffee, line 209 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5Y1fRtbry2aBzUwWV)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L209)):</sup>
Done.

---

<sup>**[setup.sh, line 24 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5TXIKgw4lqFLc6-CF)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L24)):</sup>
oops, yes!

---

<sup>**[setup.sh, line 27 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5U75VLAJXYaVE5rdH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L27)):</sup>
The problem with putting it in build/tools is that all the paths are different from that location - so it won't build there. We could separate the tools code into a special /tools directory, but then we have to have a distinct build processes for tools vs other stuff in src. I still think treating all src code uniformly for tests etc, and just having a mini build process that replicates a tiny fragment of the bigger process is the best thing we can do. 

---

<sup>**[setup.sh, line 32 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5UX743X5FZd4kjeni)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L32)):</sup>
nope, it has absolute paths. If I drop absolute paths, I would. 

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5_Ub6rZrXPvdaFLUG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L7)):</sup>
Done.

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 110 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5d1DEXQbkmYHxKu0v)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L110)):</sup>
I agree; I've made a type for the arguments and documented them in more detail there, and tried to improve names. 

What I meant by not wanting to spend a lot of time on Grunt process is w.r.t.  making a real dependency-based build system work *within grunt* - that work is bigger and I would want to drop using grunt and move to gulp for it. 

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128)



---

OK. Sorry for being picky with variable names and things but the whole team is going to be staring at this stuff for a long time so it's important to get right...

:+1:

---

<sup>**[Gruntfile.coffee, line 70 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5YZWcPvsh1YxUQ85N)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L70)):</sup>
Great. Much better.

---

<sup>**[Gruntfile.coffee, line 87 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5XYZo0lkSuL4EDyhS)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/Gruntfile.coffee#L87)):</sup>
Potentially a downside but we have/had the same problem with non-TypeScript sources too :-)

---

<sup>**[Gruntfile.coffee, line 135 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-JjBaUymuRGx1Wf2cjiW)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9b4b4acf9b5963a039036f729ce6e5ec001138c2/Gruntfile.coffee#L135)):</sup>
Oh I like this a lot better. I hadn't considered this approach. Thanks!

---

<sup>**[setup.sh, line 35 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5U75VLAJXYaVE5rdH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L35)):</sup>
Well, the files in `src/build-tools` don't depend on any non-tools code. Regarding the taskmanager test, I don't really see what taskmanager.ts gives us anyway...but that's another debate :-)

This works so I can't object too strongly but I wanted to present some alternatives.

---

<sup>**[setup.sh, line 40 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5UX743X5FZd4kjeni)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L40)):</sup>
Ah yes, I think that works out the same. Sorry!

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 114 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5d1DEXQbkmYHxKu0v)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L114)):</sup>
OK. `npmLibNames` is still pretty confusing: does it copy an entire folder from `node_modules`? It's impossible to tell from the comment.

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 120 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-JjBn7pCpwaq959cVgIy)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9b4b4acf9b5963a039036f729ce6e5ec001138c2/src/build-tools/common-grunt-rules.ts#L120)):</sup>
`pathsFromThirdPartyBuild`

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128)







---

<sup>**[setup.sh, line 35 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5U75VLAJXYaVE5rdH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L35)):</sup>
Yes, task manager is moveable, but not common grunt rules. 
Taskmanager speeds up build; used to be order of 5 or 6 times. Probably less so now. But maybe still a lot in uProxy.  

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 114 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5d1DEXQbkmYHxKu0v)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L114)):</sup>
Yeah, it's npm require magic: it's the name of the npm library. clarified in a lengthier comment. PTAL.

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 120 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-JjBn7pCpwaq959cVgIy)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9b4b4acf9b5963a039036f729ce6e5ec001138c2/src/build-tools/common-grunt-rules.ts#L120)):</sup>
Done.

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128)







---

<sup>**[setup.sh, line 35 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5U75VLAJXYaVE5rdH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/setup.sh#L35)):</sup>
So, just to point out I don't see why common-grunt-rules can't be moved out of `src`/.

---

<sup>**[src/build-tools/common-grunt-rules.ts, line 114 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128#-Jj5d1DEXQbkmYHxKu0v)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ec3480c44b943128d333219dda01c77c68d5f4b2/src/build-tools/common-grunt-rules.ts#L114)):</sup>
Ooooooh YES. Thanks.

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/128)



---

Still :+1:.

---

PR #129, opened 2015-02-27
Backport the logging changes onto v19

jpevarnek:
I think this works...

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/129)


trevj:
Nice!

:+1:

PR #130, opened 2015-02-27
Make DataChannel.getBufferedAmount() include all pending data.

bemasc:
Not really tested... but at least it builds and the existing tests pass.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/130)

### DIFF
--- a/src/webrtc/datachannel.ts
+++ b/src/webrtc/datachannel.ts
@@ -54,9 +54,13 @@ export interface DataChannel {
   // undertlying network layer for ending.
   send(data:Data) : Promise<void>;
 
+  // Returns the number of bytes which have been passed to the browser but
+  // which have not yet been handed off to usrsctplib.
+  getBrowserBufferedAmount() : Promise<number>;
+
   // Returns the number of bytes which have been passed to send() but
-  // which have not yet been transmitted to the network.
-  getBufferedAmount() : Promise<number>;
+  // which have not yet been handed off to usrsctplib.
+  getTotalBufferedAmount() : Promise<number>;
 
   // Closes this data channel.
   // A channel cannot be re-opened once this has been called.
@@ -88,6 +92,10 @@ export class DataChannelClass implements DataChannel {
   // The |toPeerDataQueue_| is chunked by the send call and conjection
   // controlled by the handler this class sets.
   private toPeerDataQueue_        :handler.Queue<Data,void>;
+  // This is the total number of bytes in all the ArrayBuffers in
+  // toPeerDataQueue_.
+  // TODO: Count bytes in strings as well.
+  private toPeerDataBytes_        :number;
 
   public onceOpened      :Promise<void>;
   public onceClosed      :Promise<void>;
@@ -105,6 +113,7 @@ export class DataChannelClass implements DataChannel {
               id:string) {
     this.dataFromPeerQueue = new handler.Queue<Data,void>();
     this.toPeerDataQueue_ = new handler.Queue<Data,void>();
+    this.toPeerDataBytes_ = 0;
 
     this.onceOpened = new Promise<void>((F,R) => {
       this.rejectOpened_ = R;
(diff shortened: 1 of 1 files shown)


iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/130)



---

This seems to change the semantics of getBufferedAmount is a way that I think is might not be as good as it can be... I think we are better to have two metrics: one for the size of stuff on `this.toPeerDataQueue_` and one for the usrsctp stack (the current 'getBufferedAmount'). Otherwise we're going to confuse ourself about what's on which bit of queuing stack. Make sense?

---

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/130)



---

Actually, neither of these measures include the usrsctp output buffer!  The current return value is the length of a queue in the browser that precedes usrsctp.

My view is that this class is an abstraction layer, and should hide these distinctions.

---

trevj:
I agree with @iislucas that there's something not quite right about collapsing the two values; if this is something callers really need, I'd be happy if it were added as a separate method (I guess this is what @iislucas is proposing too).

bemasc:
OK ... I've split the method in two.  I've renamed them both, which is safe because `getBufferedAmount` was not used anywhere.  PTAL.

I don't see what `getBrowserBufferedAmount` is useful for, but I suppose it's not expensive to keep it.

trevj:
I don't see why it's useful either! Why did you want to add this? :-)

iislucas:
OK, I think this is good; provides the basis for the instrumentation we want :+1:  from me. @trevj  ?

trevj:
Sure, :+1: . You probably want to do some version number bumping, too.

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/130)



---

OK, I've changed getTotalBufferedAmount to getJavascriptBufferedAmount, which I think is what you really wanted anyway.  PTAL.

I didn't bump package.json because we're not cutting releases from dev branch anyway.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/130)

Reviewed files:
- src/webrtc/datachannel.ts @ r7

---

Even better! :+1: :+1: 

One nit on a TODO comment. 

---

<sup>**[src/webrtc/datachannel.ts, line 97 \[r7\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/130#-JjXQt0LzxL4J9NbI1w3)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/a564a25dea79ef4ee78a931a35d1ab49c2f71800/src/webrtc/datachannel.ts#L97)):</sup>
This seems to be done below. Remote comment? or clarify you want to do it in a more accurate way? (I was pretty sure JS strings are always 2 bytes... so I think you've got it right)

---

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/130)







---

<sup>**[src/webrtc/datachannel.ts, line 97 \[r7\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/130#-JjXQt0LzxL4J9NbI1w3)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/a564a25dea79ef4ee78a931a35d1ab49c2f71800/src/webrtc/datachannel.ts#L97)):</sup>
Nope and sort-of.

`toPeerDataBytes_` only includes bytes in arraybuffers.  Strings are not counted.

Javascript uses UTF-16 (two bytes per character) to store strings ([except when it doesn't](https://blog.mozilla.org/javascript/2014/07/21/slimmer-and-faster-javascript-strings-in-firefox/)) ... but data channels use UTF-8 to send strings, and they are automatically converted by the sending and receiving browsers.

I wasn't sure what to do with this situation (what is `toPeerDataBytes_` counting anyway...), so since we don't make any intensive use of strings, I just left them out.  Suggestions welcome.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/130)







---

<sup>**[src/webrtc/datachannel.ts, line 97 \[r7\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/130#-JjXQt0LzxL4J9NbI1w3)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/a564a25dea79ef4ee78a931a35d1ab49c2f71800/src/webrtc/datachannel.ts#L97)):</sup>
I see; meh... OK, all good. :+1:

---

PR #131, opened 2015-03-02
Improvements to setup script and browserify options to make debugging easier

iislucas:
* npm dependencies are placed in their own directory (named after the npm module)
* Setup script
  * uses eval which allows && in the command. 
  * uses strict typescript (--noImplicitAllowAny)
* Allows over-riding of browserify options and makes default have an top level argument named `browserify_exports` so that debugging and interaction with top level values in a chrome interactive console is possible. 
* Exports more stuff for the logging provider top level for easier interactive debugging.

TESTED: 
 * grunt test
 * Ran sample apps 

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/131)

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -13,7 +13,7 @@ function runAndAssertCmd ()
     echo
     # We use set -e to make sure this will fail if the command returns an error
     # code.
-    set -e && $1
+    set -e && eval $1
 }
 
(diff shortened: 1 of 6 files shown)


trevj:
:+1:

PR #137, opened 2015-03-06
Mling add handlerstats

mollyling:
Queue Handler statistics.  Cheap to hold and gather, useful when we care.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/137)

### DIFF
--- a/src/handler/queue.ts
+++ b/src/handler/queue.ts
@@ -35,29 +35,23 @@ export interface QueueFeeder<Feed,Result> {
 // The |HandlerQueueStats| class contains increment-only counters
 // characterizing the state and history of a |QueueHandler|.
 export class HandlerQueueStats {
+  // Total events ever input by the HandlerQueue.  Intent:
+  // queued_events + handled_events + rejected_events = total_events.
+  total_events : number;
   // Ever-queued-events
   queued_events : number;
   // Ever-handled-evetns
   handled_events : number;
-  // Events queued after a handler was ever assigned (e.g., a handler
-  // was unset and an event was queued in the mean time.
-  following_queued_events : number;
-  // Number of events dropped in a queue clear.
-  dropped_events : number;
-
+  // Number of events rejected in a queue clear.
+  rejected_events : number;
   // Number of times a handler was set on this queue.
-  num_handlers_set : number;
+  handler_set_count : number;
   // Number of times a handler was un-set on this queue.
-  num_handlers_cleared : number;
-
-  constructor() {
-    this.queued_events = 0;
-    this.handled_events = 0;
-    this.following_queued_events = 0;
-    this.dropped_events = 0;
-    this.num_handlers_set = 0;
-    this.num_handlers_cleared = 0;
-  }
+  // handler_clear_count <= handler_set_count
+  handler_clear_count : number;
+  // Number of times we set a new handler while we have an existing
+  // promise, causing a rejection of that promise.
+  handler_rejections : number;
 }
 
 // The |QueueHandler| is the abstraction for the stream of functions that
@@ -169,7 +163,7 @@ export class Queue<Feed,Result>
   private rejectFn_ : (e:Error) => void = null;
 
   // Handler statistics.
-  private stats_  = new HandlerQueueStats;
+  private stats_  = new HandlerQueueStats();
 
   // CONSIDER: allow queue to be size-bounded? Reject on too much stuff?
   constructor() {}
@@ -188,11 +182,11 @@ export class Queue<Feed,Result>
 
   // handle or queue the given thing.
   public handle = (x:Feed) : Promise<Result> => {
-    if(this.handler_) {
+    this.stats_.total_events++;
+
+    if (this.handler_) {
       this.stats_.handled_events++;
       return this.handler_(x);
-    } else if (this.stats_.num_handlers_set > 0) {
-      this.stats_.following_queued_events++;
     }
 
     var pendingThing = new PendingPromiseHandler(x);
@@ -205,7 +199,7 @@ export class Queue<Feed,Result>
   // null. Note: a handler may itself setHandler to being null, doing so
   // should pause proccessing of the queue.
   private processQueue_ = () : void => {
-    while(this.handler_ && this.queue_.length > 0) {
+    while (this.handler_ && this.queue_.length > 0) {
       this.stats_.handled_events++;
       this.queue_.shift().handleWith(this.handler_);
     }
@@ -213,9 +207,9 @@ export class Queue<Feed,Result>
 
   // Clears the queue, and rejects all promises to handle things on the queue.
   public clear = () : void => {
-    while(this.queue_.length > 0) {
+    while (this.queue_.length > 0) {
       var pendingThing = this.queue_.shift();
-      this.stats_.dropped_events++;
+      this.stats_.rejected_events++;
       pendingThing.reject(new Error('Cleared by Handler'));
     }
   }
@@ -227,10 +221,11 @@ export class Queue<Feed,Result>
   // promise.
   public setHandler = (handler:(x:Feed) => Promise<Result>) : void => {
     if (this.rejectFn_) {
+      this.stats_.handler_rejections++;
       this.rejectFn_(new Error('Cancelled by a call to setHandler'));
       this.rejectFn_ = null;
     }
-    this.stats_.num_handlers_set++;
+    this.stats_.handler_set_count++;
     this.handler_ = handler;
     this.processQueue_();
   }
@@ -244,11 +239,12 @@ export class Queue<Feed,Result>
   // Reject the previous promise handler if it exists and stop handling stuff.
   public stopHandling = () => {
     if (this.rejectFn_) {
+      this.stats_.handler_rejections++;
       this.rejectFn_(new Error('Cancelled by a call to setHandler'));
       this.rejectFn_ = null;
     }
     this.handler_ = null;
-    this.stats_.num_handlers_cleared++;
+    this.stats_.handler_clear_count++;
   }
 
   // A convenience function that takes a T => Promise<Result> function and sets
@@ -260,14 +256,15 @@ export class Queue<Feed,Result>
   // something to handle.
   public setNextHandler = (handler:(x:Feed) => Promise<Result>)
       : Promise<Result> => {
-    this.stats_.num_handlers_set++;
+    this.stats_.handler_set_count++;
+
     return new Promise((F,R) => {
       this.setHandler((x:Feed) : Promise<Result> => {
         // Note: we don't call stopHandling() within this handler because that
         // would reject the promise we're about to fulfill.
         this.handler_ = null;
         this.rejectFn_ = null;
-        this.stats_.num_handlers_cleared++;
+        this.stats_.handler_clear_count++;
         var resultPromise = handler(x);
         resultPromise.then(F);
         return resultPromise;
--- a/src/handler/queue.ts
+++ b/src/handler/queue.ts
@@ -36,18 +36,29 @@ export interface QueueFeeder<Feed,Result> {
 // characterizing the state and history of a |QueueHandler|.
 export class HandlerQueueStats {
   // Total events ever input by the HandlerQueue.  Intent:
-  // queued_events + handled_events + rejected_events = total_events.
+  //
+  // queued_events + immediately_handled_events + rejected_events =
+  //   total_events.
+  //
+  // queued_events - queued_handled_events - rejected_events = number
+  //   of events in queue right now.
   total_events : number;
   // Ever-queued-events
   queued_events : number;
-  // Ever-handled-evetns
-  handled_events : number;
+  // Events that were immediately handled (b/c there was a handler set).
+  immediately_handled_events : number;
+  // Events that were handled after going through the queue.
+  queued_handled_events : number;
   // Number of events rejected in a queue clear.
   rejected_events : number;
-  // Number of times a handler was set on this queue.
+  // Number of times a handler was set on this queue (the handler was
+  // previously null).
   handler_set_count : number;
-  // Number of times a handler was un-set on this queue.
-  // handler_clear_count <= handler_set_count
+  // Number of times a handler was changed on this queue (the handler
+  // was previously non-null).
+  handler_change_count : number;
+  // Number of times a handler was un-set on this queue (when then
+  // handler previously non-null).
   handler_clear_count : number;
   // Number of times we set a new handler while we have an existing
   // promise, causing a rejection of that promise.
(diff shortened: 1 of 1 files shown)


iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)



---

Cool! 

A couple of questions, and a wee suggestion for alternative names and a measured event.

---

<sup>**[src/handler/queue.ts, line 36 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlEyFgrHeLfm42kuk2)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L36)):</sup>
Lets add the algebraic relationships up here, e.g. 

Let <= express that one set of events are a subset of another, and therefore should numerically be <=. The following should hold: 

`dropped_events <= queued_events`
`num_handlers_cleared <= num_handlers_set`
And if we add the suggested events below: 
`handled_events <= (on_missing_handler + on_handler_set) `
etc.

---

<sup>**[src/handler/queue.ts, line 44 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlEYfrvoOCZtHe1PZu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L44)):</sup>
I don't think I understand what `following_queued_events` is for... I think it might be better to have: 
`on_missing_handler` // events we are told to handle while no event handler is set
`on_handler_set` // events we are told to handle when an event handler is set.
`handled_events` // all events handled.
`dropped_events` // events cleared from queue
`events_rejected` // events cancelled due to a handler change.

& the handlers set and cleared as you already have. 

---

<sup>**[src/handler/queue.ts, line 194 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlC9rl0_9sWeow7ygp)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L194)):</sup>
tiny style nit: the `if` above has a return, so prefer this as just an `if` instead of `else if`, i.e. 

```

if (this.stats_.num_handlers_set > 0) {
  this.stats_.following_queued_events++;
}
```

---

<sup>**[src/handler/queue.ts, line 230 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlDD-S610yLMXRIK7C)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L230)):</sup>
Maybe add `num_rejected_handlers` ?  (will also need to be counted in stopHandling)

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)







---

<sup>**[src/handler/queue.ts, line 230 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlDD-S610yLMXRIK7C)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L230)):</sup>
actually maybe a better name: `events_rejected`

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)



---

Just one piece of important feedback, otherwise generally looks good.

---

<sup>**[src/handler/queue.ts, line 36 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlEyFgrHeLfm42kuk2)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L36)):</sup>
I really think that would be overkill. Is anyone likely to think a queue should have more dropped events than handle events?

---

<sup>**[src/handler/queue.ts, line 44 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjzBxXl5dsYNUygsS4O)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L44)):</sup>
This is the only field I'm a little wary of. It seems like you could just graph `queued_events` vs. `num_handlers_set` to see this, right? If you're maintaining a variable that can be calculated from a (very simple) combination of other variables, that's usually a smell.

---

<sup>**[src/handler/queue.ts, line 53 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjzC8FQ_A6pZOCvoHGj)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L53)):</sup>
I don't think you need the constructor: the default values will be zero (and if you do want to explicitly initialise to zero, fine, but you can just do it in the declaration).

---

<sup>**[src/handler/queue.ts, line 172 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjzBBYRNAiXV4ITFcSO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L172)):</sup>
nit: add the `()`

---

iislucas:
ping!

mollyling:
Thanks, PTAL!

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)

Reviewed files:
- src/handler/queue.ts @ r1

---

love the new names. :+1: from me. 

---

<sup>**[src/handler/queue.ts, line 39 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JkZabujzam-DyllDE15)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/842a2e3d8000f7efda5dabb49c06bfee9679bbda/src/handler/queue.ts#L39)):</sup>
This is helpful to me, thanks!

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)







---

<sup>**[src/handler/queue.ts, line 40 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-Jkco76sP2AgY7QqZUmo)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/842a2e3d8000f7efda5dabb49c06bfee9679bbda/src/handler/queue.ts#L40)):</sup>
There's something not right with this counter.

In general, counters that are the simple of two or more other counters are questionable because:
 * they waste space, as they can be computed from those other counters
 * it's surprisingly hard to get the counters exactly synchronised (exceptions can make this happen) and instrumentation is something you really want right and is easy to get right so long as you aren't trying to update one or more counters in multiple places

But this isn't even a simple sum of events -- it's conflating queuing and handling in a way I don't understand. Isn't `queued_events` + `handled_events` + `rejected_events` = `total_events * 2`?

---

<sup>**[src/handler/queue.ts, line 46 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-Jkcny_dgGuVUNxw66RO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/842a2e3d8000f7efda5dabb49c06bfee9679bbda/src/handler/queue.ts#L46)):</sup>
Suggestion: match the style of the comment to the previous counters:
```typescript
// Ever-rejected-events (due to clearing the queue or changing the handler)
```

---

<sup>**[src/handler/queue.ts, line 51 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JkcnC8dilkNtIjFL3PG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/842a2e3d8000f7efda5dabb49c06bfee9679bbda/src/handler/queue.ts#L51)):</sup>
Does the invariant still hold if I call `stopHandling` five times in a row?

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)

Reviewed files:
- src/handler/queue.ts @ r2





---

<sup>**[src/handler/queue.ts, line 40 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-Jkco76sP2AgY7QqZUmo)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/842a2e3d8000f7efda5dabb49c06bfee9679bbda/src/handler/queue.ts#L40)):</sup>
I was understanding this as at any point in time that formula holds. queued events can decrease as handled (or rejected) increases. Is that right? Needs more documentation if we're having this discussion :) 

---

mollyling:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)







---

<sup>**[src/handler/queue.ts, line 36 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlEyFgrHeLfm42kuk2)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L36)):</sup>
Some basic relationships could be helpful to make sure that we don't lose events in ways we don't understand (or forget to capture something in instrumentation).

---

<sup>**[src/handler/queue.ts, line 40 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-Jkco76sP2AgY7QqZUmo)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/842a2e3d8000f7efda5dabb49c06bfee9679bbda/src/handler/queue.ts#L40)):</sup>
Ah, yeah, I see what's going on.  There are two different aspects here:  immediate dispensation (do we queue or immediately handle?) and handling (does the event, at some point, get handled or rejected?).  The current set conflate the two.

Ok, I'll fix that up.

---

<sup>**[src/handler/queue.ts, line 44 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlEYfrvoOCZtHe1PZu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L44)):</sup>
Following on the idea of a reliable handler (which queues until an event handler is invoked), following_queued_events was supposed to be a count of events that got queued, even though we'd had a handler set at some point prior.   So, if we thought that the handler queue would always have a handler set after some initial delay, following_queued_events would always be zero.

But, as it's this confusing, I've dropped it.

---

<sup>**[src/handler/queue.ts, line 51 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JkcnC8dilkNtIjFL3PG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/842a2e3d8000f7efda5dabb49c06bfee9679bbda/src/handler/queue.ts#L51)):</sup>
It doesn't.  Yikes, does that happen?  Should it?

---

<sup>**[src/handler/queue.ts, line 189 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlC9rl0_9sWeow7ygp)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L189)):</sup>
Removed the block, after removing the counter.

---

<sup>**[src/handler/queue.ts, line 225 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlDD-S610yLMXRIK7C)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L225)):</sup>
I called it handler_rejections, to distinguish between the handler promise and each event's promise.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)







---

<sup>**[src/handler/queue.ts, line 51 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JkcnC8dilkNtIjFL3PG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/842a2e3d8000f7efda5dabb49c06bfee9679bbda/src/handler/queue.ts#L51)):</sup>
Given we track handler rejections, I'd be ok removing the handler_clear_count. Not sure it tells us a huge amount. 

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)







---

<sup>**[src/handler/queue.ts, line 36 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlEyFgrHeLfm42kuk2)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L36)):</sup>
Seem that it highlighted some bugs already; generally I am strongly in favour of making any algebra explicit. It teases out conceptual bugs and implementation ones.  We don't have to spell everything out, but key relationships help my understanding a lot. 

---

mollyling:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)



---

Clarified a bit more.

---

<sup>**[src/handler/queue.ts, line 36 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JjlEyFgrHeLfm42kuk2)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1654830c74a2b636829443bfbd4e515a6a5d219e/src/handler/queue.ts#L36)):</sup>
I teased out some of the relationships a bit more, and cleaned up what was going on.

---

<sup>**[src/handler/queue.ts, line 40 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-Jkco76sP2AgY7QqZUmo)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/842a2e3d8000f7efda5dabb49c06bfee9679bbda/src/handler/queue.ts#L40)):</sup>
Lemme know what you think.

---

<sup>**[src/handler/queue.ts, line 51 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137#-JkcnC8dilkNtIjFL3PG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/842a2e3d8000f7efda5dabb49c06bfee9679bbda/src/handler/queue.ts#L51)):</sup>
Talking to @trevj about this, I cleared up the semantics and teased out the Not-Set -> Set and Set -> Set transitions.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/137)



---

LGTM. I think this will becomes more meaningful when we use it in performance benchmarking. 

---

PR #139, opened 2015-03-10
Add the ability to set a minimum logging before messages get passed to f...

jpevarnek:
...reedom

One area we are facing a problem with performance-wise is passing
messages to freedom from the logger.  This involves sending the message
both to the core environment and then to the actual logging module.

To remove that overhead for messages we intend to do absolutely nothing
with, this adds the ability to set a minLogLevel on the the logger which
converts the message to a noop with no freedom overhead.

Adds a temporary solution for uproxy/uproxy#1005

Tested by adding a test

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/139)

### DIFF
--- a/src/logging/logging.spec.ts
+++ b/src/logging/logging.spec.ts
@@ -38,6 +38,7 @@ describe('Client logging shim using Freedom', () => {
 
   describe('Log messages', () => {
     var mockLoggerPromise :Promise<freedomTypes.Logger>;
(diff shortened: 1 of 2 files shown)


trevj:
So, we talked about this and it's only a temporary solution but there's still the problem that `minLevel` will have to be set on every logger, right? If so, I think it's probably worth deferring until we have a better solution.

PR #141, opened 2015-03-20
fix to close call of a freedom manager

iislucas:
Small fix for close arguments of a FreedomModuleFactoryManager.

tested: grunt test

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/141)


mollyling:
:+1:

PR #142, opened 2015-03-25
Add function to set minimum logging level to pass messages to freedom

jpevarnek:
One area we are facing a problem with performance-wise is passing
messages to freedom from the logger.  This involves sending the message
both to the core environment and then to the actual logging module.

To remove that overhead for messages we never intend to do anything
with, this commit adds the ability to set logging levels on the logger
to allow messages to be converted to a noop without any freedom
overhead.

Adds a temporary solution for uProxy/uproxy#1005

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/142)


jpevarnek:
Closing, going to rework again

PR #143, opened 2015-03-30
Switch Queue to use a linked list.

bemasc:
Makes handling O(1) instead of O(N), resolving https://github.com/uProxy/uproxy/issues/1178

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/143)

### DIFF
--- a/src/queue/queue.ts
+++ b/src/queue/queue.ts
@@ -1,0 +1,48 @@
+// This file defines a basic queue data structure, providing
+// the standard Array push(), shift(), and length attributes,
+// but with O(1) operation.
+
+// Private helper class.
+class Cell<T> {
+  public next :Cell<T> = null;
+
+  constructor(private item_:T) {}
+
+  public getItem() {
+    return this.item_;
+  }
+}
+
+export class Queue<T> {
+  private back_ :Cell<T> = null;
+  private front_ :Cell<T> = null;
+  public length :number = 0;
+
+  constructor() {}
+
+  // Add an item to the back of the queue.
+  public push = (item:T) : void => {
+    var cell = new Cell<T>(item);
+    if (this.length > 0) {
+      this.back_.next = cell;
+    } else {
+      // The queue was empty, so set both pointers.
+      this.front_ = cell;
+    }
+    this.back_ = cell;
+    this.length++;
+  }
+
+  // Remove and return the front element.
+  public shift = () : T => {
+    var dequeued = this.front_;
+    // If this.front_ is this.back_, then getNext() returns null.
+    this.front_ = dequeued.next;
+    dequeued.next = null;  // Just to help the garbage collector.
+    this.length--;
+    if (this.length === 0) {
+      this.back_ = null;
+    }
+    return dequeued.getItem();
+  }
+}  // class Queue
--- a/src/queue/queue.ts
+++ b/src/queue/queue.ts
@@ -5,12 +5,7 @@
 // Private helper class.
 class Cell<T> {
   public next :Cell<T> = null;
-
-  constructor(private item_:T) {}
-
-  public getItem() {
-    return this.item_;
-  }
+  constructor(public item:T) {}
 }
 
 export class Queue<T> {
@@ -43,6 +38,6 @@ export class Queue<T> {
     if (this.length === 0) {
       this.back_ = null;
     }
-    return dequeued.getItem();
+    return dequeued.item;
   }
 }  // class Queue
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -54,6 +54,7 @@ taskManager.add 'browserify_specs', [
   'browserify:loggingSpec'
   'browserify:loggingProviderSpec'
   'browserify:webrtcSpec'
+  'browserify:queueSpec'
 ]
 
 # Run unit tests
@@ -65,6 +66,7 @@ taskManager.add 'unit_test', [
   'jasmine:logging'
   'jasmine:loggingProvider'
   'jasmine:webrtc'
+  'jasmine:queue'
 ]
 
 # Run unit tests to produce coverage; these are separate from unit_tests because
@@ -77,6 +79,7 @@ taskManager.add 'coverage', [
   'jasmine:loggingCov'
   'jasmine:loggingProviderCov'
   'jasmine:webrtcCov'
+  'jasmine:queueCov'
 ]
 
 # Run unit tests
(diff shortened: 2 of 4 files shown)


iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143)



---

This is a good thing to fix! 

I would be inclined to pull out linked lists into its own data structure/class as I'm sure we'll want it in other places too - I'm sure I've done the same bad thing in other places :-)

Thoughts? 

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143)



---

Oooooooh, nice catch.

---

<sup>**[src/handler/queue.ts, line 112 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143#-JlhkZGTaMBZpBr0qgrT)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/8756c330c129af38c18bb410f843316afa4ceee4/src/handler/queue.ts#L112)):</sup>
I don't think we should be adding more public fields. Does this need to be public?

---

<sup>**[src/handler/queue.ts, line 164 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143#-JlhkeyLoe4oAnZjzkRL)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/8756c330c129af38c18bb410f843316afa4ceee4/src/handler/queue.ts#L164)):</sup>
I think we should update the comment, this is no longer a queue :-)

---

<sup>**[src/handler/queue.ts, line 207 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143#-Jlhl9IuBwbO03p6OYbK)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/8756c330c129af38c18bb410f843316afa4ceee4/src/handler/queue.ts#L207)):</sup>
I would find a comment here explaining whether the new item is added to the front or back of the queue (I'm in a meeting right now and FYI it's just a little too much mental work to figure out exactly what this is doing).

---

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143)



---

@iislucas, I don't see any other such places in the application.  Also the name of this class is "Queue", which seems sufficiently generic to me.  I'd also like to avoid "ravioli code" on this performance-critical path.

---

<sup>**[src/handler/queue.ts, line 112 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143#-JlhkZGTaMBZpBr0qgrT)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/8756c330c129af38c18bb410f843316afa4ceee4/src/handler/queue.ts#L112)):</sup>
We could wrap it with a trivial setter and getter, but I don't understand the advantage.  Note that this class is not exported, so it's only scoped to this file.

---

<sup>**[src/handler/queue.ts, line 164 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143#-JlhkeyLoe4oAnZjzkRL)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/8756c330c129af38c18bb410f843316afa4ceee4/src/handler/queue.ts#L164)):</sup>
Done.

---

<sup>**[src/handler/queue.ts, line 207 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143#-Jlhl9IuBwbO03p6OYbK)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/8756c330c129af38c18bb410f843316afa4ceee4/src/handler/queue.ts#L207)):</sup>
Done.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143)

Reviewed files:
- src/handler/queue.ts @ r1

---

@bemasc I had a look through the various repos, you're kind of right, there's only 5 other occurrences of `shift`. But that's still a few other locations where we are likely to want linked lists. 

I haven't looked in much detail, but on npm, there seem to be existing implementations also: e.g.  https://www.npmjs.com/package/linkedlist

w.r.t. The worst case performance, it looks to me like 1 extra pointer dereference, and an infrequent constant time addition for class memory allocation, and those are likely optimized away by the compiler anyway... so I'm not sure I agree with the performance beats code clarity argument for this one. At the moment, I still think we'll end up with a better library, better code quality, and code re-use if we pull out the linked list into it's own class, or use an existing implementation. Maybe I'm missing some aspect of the performance impact you have in mind? 

---

iislucas:
@bemasc: I don't want to delay work while I am out; so if you still think this is the best way forward, feel free to move forward. Or we can discuss when I am back. As you like.

bemasc:
OK, I've split the base queue out into its own class/module, preserving the Array-style API (like the linkedlist npm module).  I decided not to reuse that module because (1) it doesn't have type annotations, and (2) it uses `Object.prototype.__defineGetter__()`, which might not be fast.

> w.r.t. The worst case performance, it looks to me like 1 extra pointer dereference

For each item that goes through the queue, it's an extra object allocation and GC, plus an extra attribute-set and an extra attribute-get.  On the fast path those operations are just pointer setting and dereference, but on the slow path they are hash table operations on string keys.  Let's hope we're on the fast path.

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/143)

iislucas:
LGTM!

src/queue/queue.ts
```diff
@@ -0,0 +1,48 @@
...
+    this.back_ = cell;
+    this.length++;
+  }
+
+  // Remove and return the front element.
+  public shift = () : T => {
+    var dequeued = this.front_;
+    // If this.front_ is this.back_, then getNext() returns null.
+    this.front_ = dequeued.next;
+    dequeued.next = null;  // Just to help the garbage collector.
```
trevj: It looks to me like `dequeued.next` would *always* be null, no?

src/queue/queue.ts
```diff
@@ -0,0 +1,48 @@
...
+// the standard Array push(), shift(), and length attributes,
+// but with O(1) operation.
+
+// Private helper class.
+class Cell<T> {
+  public next :Cell<T> = null;
+
+  constructor(private item_:T) {}
+
+  public getItem() {
```
trevj: There's a weird mix of private and public fields in this file. Generally I'm against public fields but for Cell specifically I think you could make `item_` a public field.

trevj:
It seems like we need to consider renaming handler/queue.ts pretty soon! LGTM, though, and thanks for identifying and fixing.

src/queue/queue.ts
bemasc: I don't quite understand.  At the start of this line, `dequeued.next == this.front_`.

trevj: Isn't `this.front_` the only element in the queue whose `next !== null`?

bemasc: No.  `push()` sets `this.back_.next` before redefining `this.back_`.

trevj: I will trust it to you then, I just didn't want a false-sense-of-security premature optimisation creeping in :-)

src/queue/queue.ts
bemasc: Done.

src/queue/queue.ts
iislucas: I read it pretty carefully and it looks right to me.

PR #145, opened 2015-04-05
browserify for specs without coverage transform

iislucas:
While trying to do some debugging in Chrome of the SpecRunner, I noticed all specs had, by default coverage transforms applied which made them really hard to debug. So I separated coverage browserification from normal browserification. 

TESTED: 
* grunt test
* grunt coverage
* grunt dist

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/145)

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -46,7 +46,7 @@ taskManager.add 'copypasteFreedomChat', [
 ]
 
 # Create unit test code
-taskManager.add 'browserify_specs', [
+taskManager.add 'browserifySpecs', [
   'base'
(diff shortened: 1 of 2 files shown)


jpevarnek:
LGTM

---

Reviewed files:
- Gruntfile.coffee @ r1
- src/build-tools/common-grunt-rules.ts @ r1

---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/145)

trevj:
Seems like there's a prime candidate for use of a `boolean` variable in here somewhere...but otherwise :+1: :-)

PR #148, opened 2015-04-07
Update local loggers with the minimum level to send logs

jpevarnek:
This makes a few changes to how the logging system works in an effort to
reduce the number of IPC calls we make for logging.

Background:

Right now, to user the logger in a project, you load up the local logger
(logger/logger.ts) in the context of your code and then load up the
logging provider and logging controller as freedom modules.  To actually
log a message, the logger will pass the data to be logged through an IPC
to the freedom module.  Previously, this was done for every single log
message.

In a lot of cases, we would end up then discarding the log message on
the remote side due to the logging level being set at a threshold above
the message that was just logged.  In this case, we used an expensive
IPC call for no reason.

Changes:

The big change here is the creation of a new freedom interface,
LoggingListener, that will monitor the minimum level at which any action
will be taken for log messages with a specific tag.  The logger will
automatically set itself up to listen for updates to its tag if the
loggingListener module is loaded, or otherwise will function as before.

In the LoggingController module, we register each listener that is
created to its tag and will send out updates whenever its minimum level
(the lowest level between the different logging destinations where a
message would be logged) is changed.

To clean up the code a bit, the functions to change the logging
destinations have changed slightly as well.  There are now three
different functions that can be used to affect the levels:

setDefaultFilter: changes the default logging level, this is used if the
                  level is not specified for a tag

setModuleFilters: set or clear the logging levels for a list of modules.
                  Modules should be passed in as an object with the key
                  being the module name and the value being the level it
                  should be set to.  Setting the level to default clears
                  any overrides for that module.

clearFilters:     Returns all modules to use the default level filters

All three functions take the first argument as which logging destination
(current console or buffered) the level should be changed for.  All
destinations and levels are specified in enums in
loggingprovider.types.ts.

Fixes uProxy/uproxy#1005

(this message looks a lot nicer if you look at it at https://github.com/uProxy/uproxy-lib/commit/83bf7948fa3c4705a3924c75907f1ead90a5a6e5)

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/148)


trevj:
Nice. As @mollyling confirmed, this has a significant effect.

I do prefer the use of enums but since it's effectively a breaking change, you will also need to bump the major version number. Are you prepared to push this change all the way through to the UI? :-)

---



---

<sup>**[src/logging/logging.ts, line 43 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPY6jlSai5rO4db4Ku)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/logging/logging.ts#L43)):</sup>
I would comment this field. Then I think you could remove the comment in the constructor.

---

<sup>**[src/logging/logging.ts, line 50 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPYCjg4kyIyC0wzIc0)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/logging/logging.ts#L50)):</sup>
How about outputting a warning if `logginglistener` is not bound? I just ran into this gotcha.

---

<sup>**[src/logging/logging.ts, line 67 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPYNfkqFhds0eks7go)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/logging/logging.ts#L67)):</sup>
style concern: I would usually insert a blank line after these precondition-type opening statements.

---

<sup>**[src/loggingprovider/loggingprovider.ts, line 58 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPYbldYIk7Eugqej7v)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.ts#L58)):</sup>
I think complete sentences are preferable, and are more common throughout the uProxy code.

---

<sup>**[src/loggingprovider/loggingprovider.ts, line 58 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPZH5GaMTx_s7QIEaq)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.ts#L58)):</sup>
I feel this can be much simpler.

How about we add another field, to record the default level? I think this will also allow you greatly simplify `setModuleFilters`.

---

<sup>**[src/loggingprovider/loggingprovider.ts, line 151 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmP_rqq8rdxsb53PoBk)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.ts#L151)):</sup>
So, this causes log messages to appear like:
```
SocksToRtc info [Apr 8 15:37:42.500] associating session c0 with new TCP connection
```

I think I prefer the current abbreviated level, i.e.:
```
SocksToRtc I [Apr 8 15:37:42.500] associating session c0 with new TCP connection
```

(actually, outputting the level at all is kind of redundant in Chrome)

---

<sup>**[src/loggingprovider/loggingprovider.ts, line 198 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmP_6-bK8Idc1r1fPD9)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.ts#L198)):</sup>
Can we use a hash table-like stucture for lookup? It seems like it would be an easy victory.

---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
Do we need `default`?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

trevj:
If it helps, I was able to hack Simple SOCKS to avail of this with the following commit:
https://github.com/uProxy/uproxy-networking/compare/dev-requireify...trevj-logginglistener

jpevarnek:
Yup, I have the changes to `uproxy-networking` queued up in a branch, I'm thinking the changes to `uproxy` will end up being on the minimal side considering it only has the one app.

---



---

<sup>**[src/logging/logging.ts, line 43 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPY6jlSai5rO4db4Ku)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/logging/logging.ts#L43)):</sup>
Done.

---

<sup>**[src/logging/logging.ts, line 50 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPYCjg4kyIyC0wzIc0)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/logging/logging.ts#L50)):</sup>
Done.

---

<sup>**[src/logging/logging.ts, line 67 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPYNfkqFhds0eks7go)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/logging/logging.ts#L67)):</sup>
Huh, I think that not showing up might be a reviewable bug...  They show the file correctly if you only look at `r1`.

---

<sup>**[src/loggingprovider/loggingprovider.ts, line 58 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPYbldYIk7Eugqej7v)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.ts#L58)):</sup>
Done.

---

<sup>**[src/loggingprovider/loggingprovider.ts, line 151 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmP_rqq8rdxsb53PoBk)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.ts#L151)):</sup>
I have no preference between the two.  Done.

---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
It was the best way I could come up with to signify we want to clear the log level for a module

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

trevj:
---

<sup>**[src/loggingprovider/loggingprovider.ts, line 58 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPYbldYIk7Eugqej7v)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.ts#L58)):</sup>
Don't forget to do this throughout!

---

<sup>**[src/loggingprovider/loggingprovider.ts, line 153 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmP_rqq8rdxsb53PoBk)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.ts#L153)):</sup>
Awesome, thanks.

---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
Ah...what do you think of having `setModuleFilters` simply replace the current set of filters? Calling with an empty array would be equivalent to clearing all filters, and you could remove `clearFilters`, too.

(as with the optimisation to only call logginglisteners when necessary, I'm wary of introducing any sort of complexity when it has negligible benefit or use)

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

jpevarnek:
---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
That would be more consistent with the previous way of  doing things, however, being able to, as a developer, make a tweak where you decide "I'm interested in debugging-level output for this one module but would like to leave everything else the same" seems useful to me.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

trevj:
---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
Ah, yes...how about a `clearLoggingLevel(destination)`-type method then?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

jpevarnek:
---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
That would work and I would be fine moving to that if you think it is going to be significantly cleaner.  It just did not feel like it warranted an extra method when the functionality was nearly identical to setting the levels for a module.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

trevj:
---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
I think I prefer that...makes me wonder where `clearFilters` fits in?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

jpevarnek:
---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
I think having the ability to clear tag-specific levels has a potential use.

So, with this change, we would have:
`setDefaultFilter` (takes a single level)
`setModuleFilters` (takes a "dictionary")
`clearModuleFilters` (takes an array)
`clearAllFilters`

Does that sound right?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

trevj:
OK, :+1: but please think about my last comment.

You also need to bump the major version number.

---



---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
I think that would work, although I think we could get away with three methods if we were willing to make `setFilters` idempodent, e.g.:

`setDefaultFilter` (takes a single level)
`setFilters` (takes a dictionary, empty to clear)
`clearFilter`(takes a single module)

Intuitively, I prefer the idea of `setFilters` being idempodent (less code; less potential for bugs; fewer methods to maintain). However, I don't feel very strongly about it.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

jpevarnek:
---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
That sounds good to me, I'll make the change

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

jpevarnek:
---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
Oh, actually, I mistunderstood that last comment a bit.

How about:

`setDefaultFilter` (takes single level)
`setFilters` (takes a dictionary, empty to clear) (I'm interpreting what you have above as saying that it will replace all current filters)
`setFilter` (takes a single module, null value for the log level will clear)

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

trevj:
---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
Oh nice, I think that's even better!

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

jpevarnek:
---

<sup>**[src/loggingprovider/loggingprovider.types.ts, line 52 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148#-JmPT5Qzpfhm4J2wZnEH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/83bf7948fa3c4705a3924c75907f1ead90a5a6e5/src/loggingprovider/loggingprovider.types.ts#L52)):</sup>
Done.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

trevj:
Still :+1:.

---









Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/148)

mollyling:
Yay!


Cheers,
$ ls

Lally Singh | Google Ideas <http://www.google.com/ideas/> | Engineering |
lally@google.com


On Mon, Apr 13, 2015 at 2:43 PM, Jonathan Pevarnek <notifications@github.com

PR #149, opened 2015-04-09
have DataChannel block close() call until all messages have been sent

trevj:
@iislucas, since you've already examined this...

This makes `datachannel.ts` delay its call to `close()` until its `toPeerDataQueue_` is empty. I think we've been lucky until now but since @bemasc's recent backpressure change this has been an issue:
https://github.com/uProxy/uproxy-networking/pull/220

Tested primarily with this command:

```
curl -x socks5h://localhost:9999 http://www.armoryonpark.org/css/screen.css
```

Without this pull request, `curl` reports lost bytes; with, it works fine.

Fixes:
https://github.com/uProxy/uproxy/issues/1218

Does _not_ fix:
https://github.com/uProxy/uproxy/issues/1113

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/149)

PR #150, opened 2015-04-09
Make PeerConnection.close() return a Promise.

bemasc:
This is helpful for CHURN in uproxy-networking, to detect when the
browser is likely to have released the local UDP port.

This is a prerequisite for https://github.com/uProxy/uproxy-networking/pull/227

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/150)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-lib",
   "description": "Shared libraries for uProxy projects.",
-  "version": "20.1.1",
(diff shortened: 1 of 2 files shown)


trevj:
Oh, good catch, thanks. Funny, https://github.com/uProxy/uproxy-lib/pull/149 makes `DataChannel.close` return a promise :-)

:+1:

PR #153, opened 2015-04-13
Don't close the data channel until bufferedAmount goes to zero.

bemasc:
Intended to fix https://github.com/uProxy/uproxy/issues/1113

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/153)

### DIFF
--- a/src/webrtc/datachannel.spec.ts
+++ b/src/webrtc/datachannel.spec.ts
@@ -42,6 +42,7 @@ describe('DataChannel', function() {
 
     datachannel.createFromRtcDataChannel(mockRtcDataChannel).then(
         (channel:datachannel.DataChannel) => {
+      var readyToClose = false;
       channel.send({
         buffer: new Uint8Array([0, 1, 2]).buffer
       });
@@ -52,11 +53,67 @@ describe('DataChannel', function() {
       // Now that we've requested the channel be closed, "unblock" the
       // core.rtcdatachannel to drain the message queue.
       expect(sendBufferSpy).not.toHaveBeenCalled();
-      getBufferedAmountSpy.and.returnValue(Promise.resolve(0));
+      getBufferedAmountSpy.and.returnValue(Promise.resolve(
+          datachannel.PC_QUEUE_LIMIT / 2));
+
+      // Wait for 100 ms, substantially longer than the 20 ms interval used
+      // by the congestion control handler to check for room in the browser's
+      // send buffer. During that time, DataChannel should wake up, detect
+      // the free space, and call send.
+      setTimeout(() => {
+        expect(sendBufferSpy).toHaveBeenCalled();
+
+        // After send has been called, now simulate the browser successfully
+        // sending the data, and hence draining its buffer.  This means that
+        // the channel is now ready to close.
+        getBufferedAmountSpy.and.returnValue(Promise.resolve(0));
+        readyToClose = true;
+      }, 100);
 
       channel.onceClosed.then(() => {
-        expect(sendBufferSpy).toHaveBeenCalled();
+        expect(readyToClose).toBe(true);
         done();
+      });
+    });
+  });
+
+  it('the buffer loop stops if the remote peer closes', (done) => {
+    // The core.rtcdatachannel is initially open.
+    spyOn(mockRtcDataChannel, 'getReadyState').and.returnValue(
+        Promise.resolve('open'));
+
+    // Pretend the core.rtcdatachannel is buffering a lot of data.
+    var getBufferedAmountSpy = spyOn(mockRtcDataChannel, 'getBufferedAmount');
+    getBufferedAmountSpy.and.callFake(() => {
+      return Promise.resolve(datachannel.PC_QUEUE_LIMIT);
+    });
+
+    datachannel.createFromRtcDataChannel(mockRtcDataChannel).then(
+        (channel:datachannel.DataChannel) => {
+      channel.onceOpened.then(() => {
+        var baselineBufferedAmountCalls = getBufferedAmountSpy.calls.count();
+        channel.close();
+
+        // channel.close() starts the loop, which will check getBufferedAmount
+        // every 20 ms, but does not synchronously check it.
+        expect(getBufferedAmountSpy.calls.count()).toEqual(baselineBufferedAmountCalls);
+
+        // Have core.rtcdatachannel immediately emit an onclose event, as if the remote
+        // peer had closed the connection.  This should take effect in <20 ms.
+        mockRtcDataChannel.handleEvent('onclose');
+
+        channel.onceClosed.then(() => {
+          // In the next 200 ms, there should only be two additional calls to
+          // getBufferedAmount: one while checking if we are out of overflow yet,
+          // and another in the first pass of the draining loop, before that loop
+          // terminates.
+          setTimeout(() => {
+            expect(getBufferedAmountSpy.calls.count()).toEqual(
+                baselineBufferedAmountCalls + 2);
+            done();
+          }, 200);
+        });
+
       });
     });
   });
--- a/src/webrtc/datachannel.spec.ts
+++ b/src/webrtc/datachannel.spec.ts
@@ -20,6 +20,7 @@ describe('DataChannel', function() {
 
   // Ensure that close() waits for all outgoing unsent messages to be sent:
   //   https://github.com/uProxy/uproxy/issues/1218
+  //   https://github.com/uProxy/uproxy/issues/1113
   it('close waits for all data to be sent', (done) => {
     // The core.rtcdatachannel is initially open.
     spyOn(mockRtcDataChannel, 'getReadyState').and.returnValue(
--- a/src/webrtc/datachannel.ts
+++ b/src/webrtc/datachannel.ts
@@ -116,6 +116,7 @@ export class DataChannelClass implements DataChannel {
     this.fulfillDrained_ = F;
   });
 
+  // True between onceOpened and onceClosed
   private opennedSuccessfully_ :boolean;
   private rejectOpened_  :(e:Error) => void;
 
@@ -312,6 +313,11 @@ export class DataChannelClass implements DataChannel {
 
       if(bufferedAmount + CHUNK_SIZE > PC_QUEUE_LIMIT) {
         this.setOverflow_(true);
+        if (!this.opennedSuccessfully_) {
+          // The remote peer has closed the channel, so we should stop
+          // trying to drain the send buffer.
+          return;
+        }
         setTimeout(this.conjestionControlSendHandler, 20);
       } else {
         if (this.toPeerDataQueue_.getLength() === 0) {
@@ -329,12 +335,13 @@ export class DataChannelClass implements DataChannel {
     });
   }
 
+  // Closes asynchronously, after waiting for all outgoing messages.
   public close = () : Promise<void> => {
     log.debug('close requested (%1 messages and %2 bytes to send)',
-              this.toPeerDataQueue_.getLength(),
-              this.lastBrowserBufferedAmount_);
+        this.toPeerDataQueue_.getLength(),
+        this.lastBrowserBufferedAmount_);
 
-    var javascriptBufferDrained = new Promise((F, R) => {
+    var onceJavascriptBufferDrained = new Promise((F, R) => {
       if (this.getJavascriptBufferedAmount() > 0) {
         this.setOverflowListener((overflow:boolean) => {
           if (!overflow) {
@@ -347,7 +354,7 @@ export class DataChannelClass implements DataChannel {
       this.draining_ = true;
     });
 
-    javascriptBufferDrained.then(this.waitForBrowserToDrain_).then(
+    onceJavascriptBufferDrained.then(this.waitForBrowserToDrain_).then(
       this.fulfillDrained_);
 
     return this.onceClosed;
@@ -355,7 +362,8 @@ export class DataChannelClass implements DataChannel {
 
   private waitForBrowserToDrain_ = () : Promise<void> => {
     var drained :() => void;
-    var ret :Promise<void> = new Promise<void>((F, R) => {
+    var onceBrowserBufferDrained :Promise<void> =
+        new Promise<void>((F, R) => {
       drained = F;
     });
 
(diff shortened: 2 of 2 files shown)


trevj:
I patched this in Friday and it does seem to fix the issue, as described. Great work.

Two things:
 1. Can you add a unit test, or amend the existing one? Sorry, I know you're going to hate me now but I think it really helped with https://github.com/uProxy/uproxy-lib/pull/149)
 1. Can you try running the integration tests? Let's not gate merging this on the integration tests, but let's have some idea how close this helps them get to passing.

---





<sup>**[src/webrtc/datachannel.ts, line 332 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmnwYEL7nsEE7bH3cOs)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L332)):</sup>
Let's add a comment explaining that this only closes the connection once the two buffers have been emptied.

---

<sup>**[src/webrtc/datachannel.ts, line 334 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmnvocCJ71ABFsa0NTh)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L334)):</sup>
indentation: 4 spaces

---

<sup>**[src/webrtc/datachannel.ts, line 337 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmnvdzTgIwLan1zRyJj)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L337)):</sup>
I think `onceJavascriptBufferDrained` is more in line with some of our other promise variable names.

---

<sup>**[src/webrtc/datachannel.ts, line 347 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-Jmnw5hKm99GrB9ZeAtQ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L347)):</sup>
Hmm, it's harmless but you might want to set `this.draining_ = true;` at the start of this method and wrap all but the return statement in an `if (!this.draining_` block.

---

<sup>**[src/webrtc/datachannel.ts, line 358 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-Jmnvu7K121SIXztxStg)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L358)):</sup>
`onceBrowserBufferDrained`?

---

<sup>**[src/webrtc/datachannel.ts, line 366 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-Jmnwl-VKlYyBVWBwo7n)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L366)):</sup>
It seems like the value of `this.opennedSuccessfully_` will not change during the lifetime of this loop.

If so, I think that check should move outside of it?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153)

bemasc:
OK, I've extended the unit test to cover this behavior in more detail.

This doesn't seem to help much with the CHURN integration tests, unfortunately.

---





<sup>**[src/webrtc/datachannel.ts, line 332 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmnwYEL7nsEE7bH3cOs)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L332)):</sup>
Done.

---

<sup>**[src/webrtc/datachannel.ts, line 334 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmnvocCJ71ABFsa0NTh)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L334)):</sup>
Done.

---

<sup>**[src/webrtc/datachannel.ts, line 337 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmnvdzTgIwLan1zRyJj)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L337)):</sup>
Done.

---

<sup>**[src/webrtc/datachannel.ts, line 347 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-Jmnw5hKm99GrB9ZeAtQ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L347)):</sup>
The problem is, I added a check to setOverflowListener, to make it fail if you call it when this.draining_ is true (i.e. after close()).  This means that in this function, I can't set this.draining_ to true until after calling setOverflowListener.

---

<sup>**[src/webrtc/datachannel.ts, line 358 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-Jmnvu7K121SIXztxStg)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L358)):</sup>
Done.

---

<sup>**[src/webrtc/datachannel.ts, line 366 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-Jmnwl-VKlYyBVWBwo7n)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/622195b82c7e391452adb893adbb6d30e8fd37cd/src/webrtc/datachannel.ts#L366)):</sup>
This check is here because `this.opennedSuccessfully_` will transition to false if the remote peer closes the connection, and I don't want this loop to run forever in that case.

Thanks for pointing this out.  I added a unit test to make sure we don't accidentally regress on this tricky point ... and discovered that my logic was incomplete.  This loop was terminating correctly, but the congestion control loop was running forever.  I've added an additional logic change to that loop, and now the new unit test passes.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153)

trevj:
Awesome. Thanks for adding tests.

We can keep plugging away at the integration tests; I can only imagine we're getting closer.

:+1: once a couple of little remaining things are addressed.

---





<sup>**[src/webrtc/datachannel.spec.ts, line 21 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmojrTTT4Jo8lQu6scu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.spec.ts#L21)):</sup>
This test now tests *both* close cases.

That's probably okay (might be a little annoying to debug at some future point) but can you add a link in the comment to the issue you're fixing?

---

<sup>**[src/webrtc/datachannel.spec.ts, line 73 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmojfxOAfuVrf5ch3yj)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.spec.ts#L73)):</sup>
I think if you move this block inside the `setTimeout`-ed function, you could eliminate the need for `readyToClose`.

---

<sup>**[src/webrtc/datachannel.spec.ts, line 111 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmolWWmTtUsrkwZafT8)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.spec.ts#L111)):</sup>
Checking the number of calls seems orthogonal to what you're really testing -- do you think you really need it?

(if the implementation changed in such a way that the number of calls increased or decreased, would you really care?)

---

<sup>**[src/webrtc/datachannel.ts, line 120 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-Jmom5KBFCRtchB--P5P)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.ts#L120)):</sup>
This variable seems poorly named :-) Would you mind taking the opportunity to rename? It seems like `channelIsOpen_` or similar is better?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153)

bemasc:
<sup>**[src/webrtc/datachannel.spec.ts, line 21 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmojrTTT4Jo8lQu6scu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.spec.ts#L21)):</sup>
Done.

---

<sup>**[src/webrtc/datachannel.spec.ts, line 73 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmojfxOAfuVrf5ch3yj)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.spec.ts#L73)):</sup>
No, I think this separation is necessary, because the point of this section is to ensure that the test fails if onceClosed fulfills while `getBufferedAmount()` is still >0.  That means we need to call `onceClosed.then` before the channel is ready to close.

---

<sup>**[src/webrtc/datachannel.spec.ts, line 111 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmolWWmTtUsrkwZafT8)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.spec.ts#L111)):</sup>
The purpose of this test is to ensure that the number of calls after a remote close is O(1), not O(time).  Prior to a fix in revision 2 of this pull request, the number here was continuing to increment every 20ms, essentially leaking an infinite loop.

It's true that there are deeper, more precise ways to write a test like this.  Maybe the most precise way would be to mock out `setTimeout`, and then check that there are no pending events after `onceClosed`.  If you would prefer that construction, I think I can do it using `jasmine.clock().install`, but I suspect it will be extremely verbose.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153)

bemasc:
<sup>**[src/webrtc/datachannel.ts, line 120 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-Jmom5KBFCRtchB--P5P)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.ts#L120)):</sup>
Renamed to `isOpen_`.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153)

trevj:
Still :+1:, just a couple of suggestions (and suggestions only) for the tests.

---





<sup>**[src/webrtc/datachannel.spec.ts, line 74 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmojfxOAfuVrf5ch3yj)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.spec.ts#L74)):</sup>
Oh, I get it now. I don't want to spend any more time on this but I *think* in the `onceClosed` handler it would be equivalent to just check that the spy has been called. If that doesn't almost-instantly sound right to you then this is fine. Thanks!

---

<sup>**[src/webrtc/datachannel.spec.ts, line 112 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmolWWmTtUsrkwZafT8)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.spec.ts#L112)):</sup>
If it's an infinite loop, `done()` won't be called and the test would fail, right?

It's just that I would say use of `setTimeout` and checking the number of calls to methods are both slight code smells; using both in one test case is probably not ideal. But if you really think both are necessary, fine.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153)

bemasc:
<sup>**[src/webrtc/datachannel.spec.ts, line 73 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmojfxOAfuVrf5ch3yj)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.spec.ts#L73)):</sup>
I don't think it would be equivalent to check if `sendBufferSpy` or `getBufferedAmountSpy` had been called.  The equivalent thing would be to check, in `onceClosed`, that `mockRtcDataChannel.getBufferedAmount()` is returning a promise that resolves to zero.  If you think that would be clearer I can formulate it that way, but it adds another layer of asynchronicity.

---

<sup>**[src/webrtc/datachannel.spec.ts, line 112 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153#-JmolWWmTtUsrkwZafT8)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/e3a3e1811bb49c19662f3f1c0871e82c821f63e3/src/webrtc/datachannel.spec.ts#L112)):</sup>
No, `done()` will be called regardless.  What we're trying to detect is not a synchronous `while(1)` that block execution; it's a wakeup every 20ms *after* `onceClosed`.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153)

trevj:
Thanks for the explanations!

---









Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/153)

PR #157, opened 2015-04-15
terminate peerconnection when we believe it has reached an unusable state

trevj:
If we think the peerconnection has become unuseable, we should just pro-actively terminate it. This should give a better user experience when users run into the out-of-order close events issue which, sadly, seems quite easy to trigger on medium to high latency connections.

Related to this issue:
https://github.com/uProxy/uproxy/issues/1289

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/157)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-lib",
   "description": "Shared libraries for uProxy projects.",
-  "version": "24.1.0",
+  "version": "24.1.1",
   "repository": {
     "type": "git",
     "url": "https://github.com/uProxy/uproxy-lib"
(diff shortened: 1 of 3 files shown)


trevj:
Attention, @mollyling and @bemasc.

mollyling:
Worth looking into whether we can integrate reporting of these events into the rappor metrics system.

trevj:
Yes, worth having success/failure metrics in there.

iislucas:
LGTM

bemasc:
Not clear that we still want this with pool management.

trevj:
Right! Dropping this.

PR #159, opened 2015-04-21
Use circular-json to serialize objects for the log

jpevarnek:
This allows us to not fail when trying to log recursive objects.

Fixes uProxy/uproxy#1181

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/159)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,11 +17,13 @@
     }
   ],
   "peerDependencies": {
+    "circular-json": "^0.1.6",
     "freedom": "^0.6.21"
   },
   "devDependencies": {
     "arraybuffer-slice": "^0.1.2",
     "browserify-istanbul": "^0.2.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -17,11 +17,13 @@
     }
   ],
   "peerDependencies": {
+    "circular-json": "^0.1.6",
```
trevj: How come a peer dependency?

trevj:
Just one question, and this merits a patch version bump.

Otherwise, :+1:, thanks!

package.json
jpevarnek: Because we do the obnoxious copying into `build/third_party` so we need to have any node_modules we use in the project using the code.

trevj:
You forgot the patch version bump; I just republished as NPM 25.0.1.

PR #160, opened 2015-04-22
Fix handler queue stats.

bemasc:
This also includes a change to avoid calling the reject function
of the preceding NextHandler when it has no effect.

Currently all handler queue stats are `NaN` because they are initialized to `undefined`, not `0`.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/160)


mollyling:
:+1:

PR #161, opened 2015-04-22
add a heartbeat to peerconnection

trevj:
This helps Firefox detect disconnection, addressing this issue:
https://github.com/uProxy/uproxy/issues/1358

Tested with the copypaste chat sample app and Firefox (getter) with Chrome (giver).

Note that this is a _protocol breaking_ change, since old clients do not send the heartbeat.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/161)

### DIFF
--- a/src/webrtc/peerconnection.ts
+++ b/src/webrtc/peerconnection.ts
@@ -504,10 +504,9 @@ export class PeerConnectionClass implements PeerConnection<signals.Message> {
   // handled.
   private onPeerStartedDataChannel_ =
       (channelInfo:{channel:string}) : void => {
-      log.debug('%1: peer created channel %2',
-          this.peerName_, channelInfo.channel);
       this.addRtcDataChannel_(channelInfo.channel).then((dc:DataChannel) => {
         var label :string = dc.getLabel();
+        log.debug('%1: peer created channel %2', this.peerName_, label);
         if (label === PeerConnectionClass.CONTROL_CHANNEL_LABEL) {
           // If the peer has started the control channel, register it
           // as this user's control channel as well.
--- a/src/webrtc/peerconnection.ts
+++ b/src/webrtc/peerconnection.ts
@@ -106,10 +106,10 @@ export class PeerConnectionClass implements PeerConnection<signals.Message> {
 
   // Interval, in milliseconds, after which the peerconnection will
   // terminate if no heartbeat is received from the peer.
-  private static HEARTBEAT_TIMEOUT_MS_ = 10000;
+  private static HEARTBEAT_TIMEOUT_MS_ = 30000;
 
   // Interval, in milliseconds, at which heartbeats are sent to the peer.
-  private static HEARTBEAT_INTERVAL_MS_ = 2000;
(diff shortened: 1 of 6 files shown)


trevj:
With sadness.

@dborkan as-is, this is a protocol breaking change...will that be a show-stopper for the frontend?

trevj:
Please note that this includes a few commits which improve peerconnection logging; that was part of a larger effort to tidy that class' code. Since it greatly helps debug this change, I decided to leave it in. Apologies for conflating pull requests.

src/webrtc/peerconnection.ts
```diff
@@ -104,6 +104,16 @@ export interface PeerConnection<TSignallingMessage> {
 //      3.1. completeConnection_ -> [Fulfill onceConnected]
 export class PeerConnectionClass implements PeerConnection<signals.Message> {
 
+  // Interval, in milliseconds, after which the peerconnection will
+  // terminate if no heartbeat is received from the peer.
+  private static HEARTBEAT_TIMEOUT_MS_ = 10000;
```
bemasc: I think this is very pessimistic.  TCP sockets will survive a 2 minute disconnection.  This means a 10 second network interruption will always kill your proxy session.

dborkan:
Protocol breaking changes are OK with me before public launch.  Our current
batch of closed-beta testers have been having a lot of issues, plus I think
we already did a protocol breaking change just yesterday by publishing the
version to the webstore which removes datachannel closing.

src/webrtc/peerconnection.ts
```diff
@@ -535,19 +538,78 @@ export class PeerConnectionClass implements PeerConnection<signals.Message> {
...
+            this.peerName_, e.message);
+      });
+
+      if (Date.now() - lastPingTimestamp >
+          PeerConnectionClass.HEARTBEAT_TIMEOUT_MS_) {
+        log.debug('%1: heartbeat timeout, terminating', this.peerName_);
+        this.closeWithError_('no heartbeat received for >' +
+            PeerConnectionClass.HEARTBEAT_TIMEOUT_MS_ + 'ms');
+      } else if (!terminate) {
+        setTimeout(send, PeerConnectionClass.HEARTBEAT_INTERVAL_MS_);
```
bemasc: I think this would be clearer with `setInterval`.

iislucas:
+1 that is my view too.

src/webrtc/peerconnection.ts
trevj: Yes, thanks!

src/webrtc/peerconnection.ts
trevj: Yup. Two minutes is a long time for your browser to be unresponsive with no feedback from uProxy...how about 30 seconds?

bemasc: OK, let's try 30 seconds.

trevj: OK, I bumped this to 30s and changed the interval to 5s.

trevj:
Thanks for looking at this so quickly!

bemasc:
:+1:

PR #163, opened 2015-04-24
add module names to LoggingController.getLogs

trevj:
When outputting to the console we go via the freedomjs provider, which pretty prints metadata such as the module name; so, when, exporting to strings we should export all the fields.

Tested with Simple SOCKS and uProxy.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/163)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-lib",
   "description": "Shared libraries for uProxy projects.",
-  "version": "25.0.1",
+  "version": "25.0.2",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


jpevarnek:
LGTM

---

Reviewed files:
- package.json @ r1
- src/loggingprovider/loggingprovider.ts @ r1

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/163)

trevj:
Fixes this issue:
https://github.com/uProxy/uproxy/issues/1291

trevj:
Thanks!

PR #164, opened 2015-04-29
Add typing from freedomjs-anonymized-metrics

dborkan:
Add typing from freedomjs-anonymized-metrics (https://github.com/willscott/freedomjs-anonymized-metrics/blob/master/anonmetrics.json)

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/164)


trevj:
Cool!

This raises an interesting question about where to place typings for third-party freedomjs modules. I think this may be the first example. I think I would prefer to place this in a separate folder, e.g. like what we did here for various crypto-related NPMs:
https://github.com/uProxy/uproxy-lib/blob/dev/src/crypto/random.ts

Also, is this likely to be used outside the frontend? It may be more appropriate in the uproxy repo.

---





<sup>**[package.json, line 4 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164#-JoUH8fDdPHVBya5Ms-r)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/cda1276403f560c320380afbd153408638342783/package.json#L4)):</sup>
Since this is a new API, bumping the middle (minor) number is more appropriate.

---

<sup>**[third_party/freedom-typings/anonymized-metrics.d.ts, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164#-JoUHLJa2xQOdSpe2Yex)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/cda1276403f560c320380afbd153408638342783/third_party/freedom-typings/anonymized-metrics.d.ts#L7)):</sup>
I believe it's better to use `Object` rather than any, when possible.

---

<sup>**[third_party/freedom-typings/anonymized-metrics.d.ts, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164#-JoUHDRh9aLI3iAuytcI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/cda1276403f560c320380afbd153408638342783/third_party/freedom-typings/anonymized-metrics.d.ts#L9)):</sup>
Since this is a freedomjs module, there are also the `on` messages to consider. Here's a good example from uproxy-networking:
https://github.com/uProxy/uproxy-networking/blob/dev/src/churn-pipe/freedom-module.interface.ts#L23

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164)

dborkan:
PTAL.  I think that eventually this will be used in uproxy's networking code (reporting metrics on number of sockets, etc).  Also freedom-typings seems like the sane place to me to have this - it's similar to other .d.ts files we have for the social interface, tcpsocket, etc

---





<sup>**[third_party/freedom-typings/anonymized-metrics.d.ts, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164#-JoUHLJa2xQOdSpe2Yex)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/cda1276403f560c320380afbd153408638342783/third_party/freedom-typings/anonymized-metrics.d.ts#L7)):</sup>
It really is an any type - right now numbers and strings are both supported

---

<sup>**[third_party/freedom-typings/anonymized-metrics.d.ts, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164#-JoUHDRh9aLI3iAuytcI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/cda1276403f560c320380afbd153408638342783/third_party/freedom-typings/anonymized-metrics.d.ts#L9)):</sup>
That one doesn't have on or emit either.  Currently anonymized-metrics doesn't have any events it raises so I'm not sure on is needed here

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164)

trevj:
As discussed, let's move the new typing under `third_party/freedom-anonymized-metrics/.

---





<sup>**[third_party/freedom-typings/anonymized-metrics.d.ts, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164#-JoUHLJa2xQOdSpe2Yex)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/cda1276403f560c320380afbd153408638342783/third_party/freedom-typings/anonymized-metrics.d.ts#L7)):</sup>
Seems we can use the union type: `number|string`

---

<sup>**[third_party/freedom-typings/anonymized-metrics.d.ts, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164#-JoUHDRh9aLI3iAuytcI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/cda1276403f560c320380afbd153408638342783/third_party/freedom-typings/anonymized-metrics.d.ts#L9)):</sup>
OK. If we don't use it and there won't be any, cool.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164)

dborkan:
Just changed to string|number and also moved the file, PTAL, thanks

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164)

trevj:
Path for new file: `third_party/freedomjs-anonymized-metrics/freedomjs-anonymized-metrics.d.ts`

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164)

dborkan:
Just moved the file, PTAL

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164)

trevj:
:+1:

Thanks for bearing with me!

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/164)

PR #165, opened 2015-05-05
Reduce frequency of calls to getBufferedAmount

bemasc:
This appears to increase the throughput of proxying by 33 percent when CPU-limited (equivalent to a 25% reduction in CPU intensity).  This is consistent with performance being IPC-dominated, because this change reduces the number of IPC messages per forwarded block from 5 (TCP 'message' event, getBufferedAmount, response, send, response) to 3 (a 40% reduction).

Extrapolating from these numbers, the remaining 3 IPCs account for 50% of CPU time after this change, so max throughput would double if they could be removed.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/165)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-lib",
   "description": "Shared libraries for uProxy projects.",
-  "version": "25.0.3",
+  "version": "25.0.4",
   "repository": {
     "type": "git",
     "url": "https://github.com/uProxy/uproxy-lib"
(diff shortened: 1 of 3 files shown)


trevj:
NICE!

On my system this bumps download speed from ~2500Kb/sec to ~3300Kb/sec.

Can you bump the patch version? Then, huge :+1:.

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/165)

PR #167, opened 2015-05-08
Discard pending inbound data when the local side closes a socket.

bemasc:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/167)


trevj:
Great! :+1:

PR #170, opened 2015-05-08
Add missing copypaste messages to non-English locales.

jetpack:
I don't know Italian, Arabic, or Farsi sadly, so I just copied the English
messages, but I think it's better than blanks. Also did some whitespace
formatting to make it easier to diff and see which messages are missing.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/170)


iislucas:
(for reference, looks like this was replaced by https://github.com/uProxy/uproxy-lib/pull/171)

PR #171, opened 2015-05-08
Add missing copypaste messages to non-English locales.

jetpack:
Add missing copypaste messages to non-English locales.

I don't know Italian, Arabic, or Farsi sadly, so I just copied the English messages, but I think it's better than blanks.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/171)


iislucas:
Thanks for the fixes!

Maybe just run the text fields through google translate for now? 

And and add a note on how you tested.

trevj:
Agreed, Google Translate is where most of the current translations came from. Thanks for adding these!

---





<sup>**[src/samples/copypaste-socks-chromeapp/locales/ar/messages.json, line 6 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/171#-Jp2OhfHSZvhsrlowt1b)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/36f775cbf3a890608155f00883468a205f232688/src/samples/copypaste-socks-chromeapp/locales/ar/messages.json#L6)):</sup>
Now I'm curious...@lucyhe: if omitted, will a key default to the English text?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/171)

jetpack:
Used Google Translate for the new messages. Also, I think "SOCKS" was being incorrectly translated as "socks" (e.g., stockings) for Arabic. Fixed that.

@trevj Nope, a missing key results in an empty string (which is how I noticed some missing messages in the first place, when I switched to Italian on a whim).

Tested manually by checking messages appear for each language. Previously, messages were blank.

jetpack:
ping?

trevj:
Thanks again for doing this!

:+1:

PR #175, opened 2015-05-12
Increase datachannel buffer limit to 16 KiB.

bemasc:
This seems to be approximately the largest (most efficient) value
that is stable under extreme loads on a slow test machine.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/175)


trevj:
Awesome. This seems good enough for now. :+1:.

PR #176, opened 2015-05-13
fix send-on-close test

trevj:
This should fix the Shippable tests. I guess this got broken with some of the recent changes to when we check the brower's buffer size; I don't think anything was really broken, just the test.

I can't remember why set the buffered size to half the maximum amount mid-way through the test, can you?

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/176)


bemasc:
:+1: 
The purpose of setting the buffered amount to half the maximum was to check that `DataChannel` doesn't wait for `bufferedAmount` to reach zero before resuming transmission.

PR #180, opened 2015-05-22
oops, test should include integration tests

trevj:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/180)


salomegeo:
won't travis try to run integration tests as well and fail because it has no chrome?

PR #182, opened 2015-05-27
Use reckless calls where possible, for performance

bemasc:
Use reckless calls where possible for performance.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/182)

### DIFF
--- a/src/churn-pipe/churn-pipe.ts
+++ b/src/churn-pipe/churn-pipe.ts
@@ -194,14 +194,12 @@ class Pipe {
    * Sends a message over the network to the specified destination.
    * The message is obfuscated before it hits the wire.
    */
-  private sendTo_ = (buffer:ArrayBuffer, to:net.Endpoint) => {
+  private sendTo_ = (buffer:ArrayBuffer, to:net.Endpoint) : void => {
     var transformedBuffer = this.transformer_.transform(buffer);
-    return this.publicSocket_.sendTo(
+    this.publicSocket_.sendTo.reckless(
       transformedBuffer,
       to.address,
-      to.port).then(() => {
-        return Promise.resolve();
-      });
+      to.port)
   }
 
   /**
@@ -217,7 +215,7 @@ class Pipe {
       port: recvFromInfo.port
     };
     this.getMirrorSocket_(source).then((mirrorSocket:freedom_UdpSocket.Socket) => {
-      mirrorSocket.sendTo(
+      mirrorSocket.sendTo.reckless(
           buffer,
           this.browserEndpoint_.address,
           this.browserEndpoint_.port);
--- a/src/churn-pipe/churn-pipe.ts
+++ b/src/churn-pipe/churn-pipe.ts
@@ -199,7 +199,7 @@ class Pipe {
     this.publicSocket_.sendTo.reckless(
       transformedBuffer,
       to.address,
-      to.port)
+      to.port);
   }
 
   /**
--- a/src/net/tcp.ts
+++ b/src/net/tcp.ts
@@ -430,11 +430,11 @@ export class Connection {
   }
 
   public pause = () => {
-    this.counter_.wrap(this.connectionSocket_.pause);
+    this.connectionSocket_.pause.reckless();
   }
 
   public resume = () => {
-    this.counter_.wrap(this.connectionSocket_.resume);
+    this.connectionSocket_.resume.reckless();
   }
 
   // This is called to close the underlying socket. This fulfills the
--- a/src/net/tcp.ts
+++ b/src/net/tcp.ts
@@ -368,9 +368,8 @@ export class Connection {
     // |dataToSocketQueue| allows a class using this connection to start
     // queuing data to be send to the socket.
     this.onceConnected.then(() => {
-      this.dataToSocketQueue.setSyncHandler((buffer:ArrayBuffer) : void => {
-        this.connectionSocket_.write.reckless(buffer);
-      });
+      this.dataToSocketQueue.setSyncHandler(
+          this.connectionSocket_.write.reckless);
     });
     this.onceConnected.catch((e:Error) => {
       this.fulfillClosed_(SocketCloseKind.NEVER_CONNECTED);
@@ -474,6 +473,8 @@ export class Connection {
    * Sends a message that is pre-formatted as an arrayBuffer.
    */
   public send = (msg :ArrayBuffer) : Promise<void> => {
+    // This will reject if the socket is closed before the
+    // data can be sent.
     return this.dataToSocketQueue.handle(msg);
   }
 
--- a/src/net/tcp.ts
+++ b/src/net/tcp.ts
@@ -368,9 +368,8 @@ export class Connection {
     // |dataToSocketQueue| allows a class using this connection to start
     // queuing data to be send to the socket.
     this.onceConnected.then(() => {
-      this.dataToSocketQueue.setHandler((buffer:ArrayBuffer) => {
+      this.dataToSocketQueue.setSyncHandler((buffer:ArrayBuffer) : void => {
         this.connectionSocket_.write.reckless(buffer);
-        return Promise.resolve<void>();
       });
     });
     this.onceConnected.catch((e:Error) => {
--- a/src/net/tcp.ts
+++ b/src/net/tcp.ts
@@ -271,7 +271,7 @@ export class Connection {
   // Queue of data to be handled, and the capacity to set a handler and
   // handle the data.
   public dataFromSocketQueue :handler.Queue<ArrayBuffer,void>;
-  public dataToSocketQueue :handler.Queue<ArrayBuffer, freedom_TcpSocket.WriteInfo>;
+  public dataToSocketQueue :handler.Queue<ArrayBuffer,void>;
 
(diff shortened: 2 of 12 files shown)


src/churn-pipe/churn-pipe.ts
```diff
@@ -194,14 +194,12 @@ class Pipe {
...
+  private sendTo_ = (buffer:ArrayBuffer, to:net.Endpoint) : void => {
     var transformedBuffer = this.transformer_.transform(buffer);
-    return this.publicSocket_.sendTo(
+    this.publicSocket_.sendTo.reckless(
       transformedBuffer,
       to.address,
-      to.port).then(() => {
-        return Promise.resolve();
-      });
+      to.port)
```
trevj: missing semi-colon

src/net/tcp.ts
```diff
@@ -370,8 +369,8 @@ export class Connection {
     // queuing data to be send to the socket.
     this.onceConnected.then(() => {
       this.dataToSocketQueue.setHandler((buffer:ArrayBuffer) => {
-        return this.counter_.wrap(this.connectionSocket_.write.bind(
-                this.connectionSocket_, buffer));
+        this.connectionSocket_.write.reckless(buffer);
+        return Promise.resolve<void>();
```
trevj: can't we use `setSyncHandler`, then we don't have to return a dummy promise?

third_party/freedom-typings/freedom-common.d.ts
```diff
@@ -134,6 +134,22 @@ declare module freedom {
     // TODO: explore how to use FreedomModuleFactoryManager.
     [moduleName:string] : FreedomModuleFactoryManager<any>;
   }
+
+  // This generic interface represents any freedom method. Its purpose is to extend
+  // the basic definition to include the reckless call method, which does not
+  // produce a reply message.
+  interface Method1<T,R> {
```
trevj: the number suffix reminds me of a certain Java API i've worked with in the past...

seems okay, but how about `Reckless1`, `Reckless2`, ...?

third_party/freedom-typings/rtcdatachannel.d.ts
```diff
@@ -34,8 +34,8 @@ declare module freedom_RTCDataChannel {
     close() : Promise<void>;
     getBinaryType() : Promise<string>;
     setBinaryType(type:string) : Promise<void>;
-    send(message:string) : Promise<void>;
-    sendBuffer(message:ArrayBuffer) : Promise<void>;
+    send :freedom.Method1<string,void>;
```
trevj: so, do all void-returning methods have a reckless counterpart?

trevj:
Great. Thanks for pushing this through on the freedomjs side.

Do you have a (very is fine) rough idea of the throughput boost this provides?

src/churn-pipe/churn-pipe.ts
bemasc: Done.

src/net/tcp.ts
bemasc: Good point! Done.

third_party/freedom-typings/freedom-common.d.ts
bemasc: Yeah, `Method` is not a great name ... but I'm not sure `Reckless` is right either, because _every_ method supports this pattern.

third_party/freedom-typings/rtcdatachannel.d.ts
bemasc: No, all methods have a void-returning reckless counterpart :-).

bemasc:
Just tested, seeing 10% boost on CHURN, or maybe a hair more.

third_party/freedom-typings/freedom-common.d.ts
trevj: Right, since it applies to every function I guess this makes sense.

third_party/freedom-typings/rtcdatachannel.d.ts
trevj: Cool, got it.

trevj:
:+1:

bemasc:
I extended this change to include pause() and resume().  Now seeing about a 15% boost.

src/net/udpclient.ts
```diff
@@ -36,7 +37,20 @@ class UdpClient {
   public bind() : Promise<net.Endpoint> {
     // TODO: not sure what else this should be?
     return this.socket.bind('127.0.0.1', 0)
-        .then((endpoint:net.Endpoint) => {
+        .then((resultCode:number) => {
```
trevj: This file has been deleted, so you might want to re-base :-)

Sorry I didn't do this earlier...

src/net/tcp.ts
```diff
@@ -369,9 +368,8 @@ export class Connection {
     // |dataToSocketQueue| allows a class using this connection to start
     // queuing data to be send to the socket.
     this.onceConnected.then(() => {
-      this.dataToSocketQueue.setHandler((buffer:ArrayBuffer) => {
-        return this.counter_.wrap(this.connectionSocket_.write.bind(
-                this.connectionSocket_, buffer));
+      this.dataToSocketQueue.setSyncHandler((buffer:ArrayBuffer) : void => {
+        this.connectionSocket_.write.reckless(buffer);
```
trevj: seems like a candidate for `bind` but obviously this is fine

src/net/tcp.ts
```diff
@@ -475,7 +473,7 @@ export class Connection {
   /**
    * Sends a message that is pre-formatted as an arrayBuffer.
    */
-  public send = (msg :ArrayBuffer) : Promise<freedom_TcpSocket.WriteInfo> => {
+  public send = (msg :ArrayBuffer) : Promise<void> => {
```
trevj: is there any need to return anything here now?

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -474,24 +474,21 @@ import logging = require('../logging/logging');
...
       if (!data.buffer) {
-        return Promise.reject(new Error(
-            'received non-buffer data from datachannel'));
+        throw new Error('received non-buffer data from datachannel');
       }
       this.bytesReceivedFromPeer_.handle(data.buffer.byteLength);
       this.channelReceivedBytes_ += data.buffer.byteLength;
 
-      return this.tcpConnection_.send(data.buffer);
+      this.tcpConnection_.send(data.buffer);
```
trevj: question: what happens when a reckless call raises an exception?

src/webrtc/datachannel.ts
```diff
@@ -285,28 +285,25 @@ export class DataChannelClass implements DataChannel {
...
         this.lastBrowserBufferedAmount_ += data.buffer.byteLength;
       } else {
         // Data is good when it meets the type expected of the Data. If type-
         // saftey is ensured at compile time, this should never happen.
-        return Promise.reject(new Error(
-            'Bad data: ' + JSON.stringify(data)));
+        throw new Error('Bad data: ' + JSON.stringify(data));
       }
     // Can raise NetworkError if channel died, for example.
     } catch (e) {
```
trevj: i just noticed this nasty "log and re-throw" anti-pattern -- might be an opportunity now to remove it?

trevj:
Looking very very good. Just a couple of things...mostly, I'm wondering how exceptions behave with reckless calls.

src/net/udpclient.ts
bemasc: Thanks! Done.

src/net/tcp.ts
bemasc: I added a comment here noting that the Promise can reject under certain circumstances.  I think that might make the return worthwhile.

src/net/tcp.ts
bemasc: Done.  (Doesn't even need bind, actually...)

src/rtc-to-net/rtc-to-net.ts
bemasc: Reckless calls never raise an exception (that's why they're reckless!).

If the method call throws or rejects in the core environment, it should trigger a freedom log message, so we should still see it in the logs.

src/webrtc/datachannel.ts
bemasc: Done.

src/churn-pipe/churn-pipe.ts
trevj: :+1:

src/net/tcp.ts
trevj: :+1:

third_party/freedom-typings/freedom-common.d.ts
trevj: :+1:

third_party/freedom-typings/rtcdatachannel.d.ts
trevj: :+1:

src/rtc-to-net/rtc-to-net.ts
trevj: OK. That seems like something we might want to change about `reckless`.

trevj:
:+1:

PR #184, opened 2015-05-28
simple socks should not specify STUN servers

trevj:
Works around for this issue (also, there is just no need for these samples to specify STUN servers):
https://github.com/uProxy/uproxy/issues/1591

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/184)


bemasc:
:+1:

PR #185, opened 2015-05-29
fallback to Math.random for key generation on firefox

trevj:
churn has some issues on Firefox. This is the easy fix, working around this issue:
https://github.com/uProxy/uproxy/issues/1593

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/185)

### DIFF
--- a/src/churn/churn.ts
+++ b/src/churn/churn.ts
@@ -174,7 +174,7 @@ var log :logging.Log = new logging.Log('churn');
   // Generates a key suitable for use with CaesarCipher, viz. 1-255.
   var generateCaesarKey_ = (): number => {
     try {
-      return (random.randomUint32() % 254) + 1;
+      return (random.randomUint32() % 255) + 1;
     } catch (e) {
       // https://github.com/uProxy/uproxy/issues/1593
       log.debug('crypto unavailable, using Math.random');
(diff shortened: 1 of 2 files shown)


src/churn/churn.ts
```diff
@@ -172,8 +172,14 @@ var log :logging.Log = new logging.Log('churn');
   };
 
   // Generates a key suitable for use with CaesarCipher, viz. 1-255.
-  var generateCaesarKey_ = () : number => {
-    return (random.randomUint32() % 254) + 1;
+  var generateCaesarKey_ = (): number => {
+    try {
+      return (random.randomUint32() % 254) + 1;
```
bemasc: I think you actually want % 255 here.  (255 % 255 == 0)

bemasc:
:+1:

willscott:
Keys should never be math.random. that already got flagged as a security
issue in our first review.

bemasc:
This is a Caesar Cipher.  Using math.random hardly seems like the main cryptographic vulnerability.

trevj:
Right, calling this a "key" is generous -- it just ensures we don't have a single, easily blockable signature. A very slight advance over a fixed "key".

src/churn/churn.ts
trevj: but we want a number from 1-255, so generate a number between 0-254, then add 1

(right?)

willscott: `random % 254` will give you between 0-253.

trevj: Oops. Thanks!

PR #187, opened 2015-06-02
remove ancient, unused udp-for-socks classes

trevj:
Random cleanup. This hasn't been used since we dropped UDP support in the SOCKS proxy. We may want that again but the code will remain in git history (doing so was the only reason I didn't delete these classes when merging in uproxy-networking's code).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/187)


bemasc:
:+1:

PR #188, opened 2015-06-02
remove pcState and dataChannels public fields from PeerConnection interface

trevj:
Addresses:
https://github.com/uProxy/uproxy/issues/583

Tested with Simple SOCKS and by compiling uProxy against my client; these fields aren't used anywhere.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/188)


bemasc:
:+1:

PR #189, opened 2015-06-02
remove onceConnecting public field

trevj:
More public field removal: `onceConnecting` isn't used anywhere.

Also, remove references to `pcState` in various comments.

Tested with copypaste and uProxy.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/189)


bemasc:
:+1:

PR #194, opened 2015-06-09
do set set remote ICE candidates until setRemoteDescription has resolved

trevj:
Fixes:
https://github.com/uProxy/uproxy/issues/1623

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/194)

### DIFF
--- a/src/webrtc/peerconnection.spec.ts
+++ b/src/webrtc/peerconnection.spec.ts
@@ -10,9 +10,6 @@ import RTCPeerConnection = freedom_RTCPeerConnection.RTCPeerConnection;
 import RTCDataChannelInit = freedom_RTCPeerConnection.RTCDataChannelInit;
 
 import freedomMocker = require('../freedom/mocks/mock-freedom-in-module-env');
-freedom = freedomMocker.makeMockFreedomInModuleEnv({
-  'rtcdatachannel': () => { return new MockFreedomRtcDataChannel(); }
-});
 
 import signals = require('./signals');
 import peerconnection = require('./peerconnection');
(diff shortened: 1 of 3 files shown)


src/webrtc/peerconnection.ts
```diff
@@ -317,47 +325,47 @@ export class PeerConnectionClass implements PeerConnection<signals.Message> {
...
   }
 
   private handleCandidateSignalMessage_ = (
-      candidate:freedom_RTCPeerConnection.RTCIceCandidate) : void => {
+      candidate:freedom_RTCPeerConnection.RTCIceCandidate)
+      : Promise<void> => {
     // CONSIDER: Should we be passing/getting the SDP line index?
     // e.g. https://code.google.com/p/webrtc/source/browse/stable/samples/js/apprtc/js/main.js#331
-    this.pc_.addIceCandidate(candidate).catch((e: Error) => {
+    return this.pc_.addIceCandidate(candidate).catch((e: Error) => {
```
bemasc: Is anything using this return value?

trevj: No. I just figured since `addIceCandidate` *does* return a promise that it was no harm.

src/webrtc/peerconnection.spec.ts
```diff
@@ -81,22 +77,59 @@ describe('PeerConnection', function() {
...
+    var setRemoteDescriptionSpy = spyOn(mockRtcPeerConnection,
+        'setRemoteDescription').and.callFake(
+        (desc:freedom_RTCPeerConnection.RTCSessionDescription) => {
+      pc.handleSignalMessage({
+        type: signals.Type.CANDIDATE,
+        candidate: {
+          candidate: 'fakeCandidate'
+        }
+      });
+      expect(mockRtcPeerConnection.addIceCandidate).not.toHaveBeenCalled();
```
bemasc: I think it would be clearer to move this check to `MockRtcPeerConnection.addIceCandidate`.

src/webrtc/peerconnection.ts
```diff
@@ -381,21 +389,25 @@ export class PeerConnectionClass implements PeerConnection<signals.Message> {
...
-        this.handleAnswerSignalMessage_(message.description);
+        this.handleAnswerSignalMessage_(message.description).then(
+            this.fulfillHaveRemoteDescription_);
         break;
       // Add remote ice candidate.
       case signals.Type.CANDIDATE:
-        this.handleCandidateSignalMessage_(message.candidate);
+        this.onceHaveRemoteDescription_.then(() => {
+          this.handleCandidateSignalMessage_(message.candidate);
+        });
```
bemasc: Might as well add a `.catch` here, now that `handleCandidateSignalMessage_` returns a Promise.

src/webrtc/peerconnection.spec.ts
trevj: As discussed, I added a check that `addIceCandidate` *is* called.

This required a bit of refactoring. PTAL!

src/webrtc/peerconnection.ts
trevj: Right, but `onceHaveRemoteDescription` never throws (we never call the rejection function).

bemasc:
:+1:

PR #196, opened 2015-06-10
have mirror sockets bind to 0.0.0.0 and accept incoming packets only from the browser

trevj:
Tested with uProxy, one peer running Chrome and the other Firefox.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/196)

### DIFF
--- a/src/churn-pipe/churn-pipe.ts
+++ b/src/churn-pipe/churn-pipe.ts
@@ -174,7 +174,24 @@ class Pipe {
         throw new Error('bindRemote failed with result code ' + resultCode);
       }
       mirrorSocket.on('onData', (recvFromInfo:freedom_UdpSocket.RecvFromInfo) => {
-        this.sendTo_(recvFromInfo.data, remoteEndpoint);
+        // Ignore packets that do not originate from the browser, for a
+        // theoretical security benefit.
+        if (!this.browserEndpoint_) {
+          log.warn('browser endpoint not set, mirror socket for %1 ignoring ' +
+              'incoming packet from %2', remoteEndpoint, {
+                address: recvFromInfo.address,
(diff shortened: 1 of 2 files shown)


bemasc:
Isn't this unsafe, without any source filtering?

trevj:
Just doing one thing at a time.

bemasc:
Introducing a known security-ish issue into a point release, with the intention of fixing it in the following one, makes me very uncomfortable.  I would prefer to do these changes in the other order.

trevj:
This was a lot easier than I expected!

PTAL. Tested with uProxy between Chrome and Firefox and Simple SOCKS on both Chrome and Firefox and manually sending packets to a mirror socket with ncat.

src/churn-pipe/churn-pipe.ts
```diff
@@ -164,13 +164,34 @@ class Pipe {
...
-        this.sendTo_(recvFromInfo.data, remoteEndpoint);
+        // Ignore packets that do not originate from the browser, for a
+        // theoretical security benefit.
+        if (!this.browserEndpoint_) {
+          log.warn('browser endpoint not set, mirror socket for %1 ignoring ' +
+              'incoming packet from %2', remoteEndpoint, {
+                address: recvFromInfo.address,
+                port: recvFromInfo.port
+              });
+        } else if (recvFromInfo.address !== this.browserEndpoint_.address &&
```
bemasc: `||`

bemasc:
:+1:

src/churn-pipe/churn-pipe.ts
trevj: Oops! Fixed. Thanks.

PR #197, opened 2015-06-10
no need to send churn candidates to the remote peer

trevj:
I made this change when debugging the recent localhost-related issue, thinking it might be why Firefox was having trouble parsing candidates. It's not, but I kinda like how it eliminates the need to send a bunch of (useless) candidates to the remote peer.

Thoughts?

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/197)


bemasc:
:+1:

PR #198, opened 2015-06-11
eliminate errors on churn startup, with some other logging cleanup

trevj:
Mostly, this eliminates the nasty uncaught exception for every churn connection.

Also, a few other bits and bobs I would have appreciated recently when debugging churn. We have been logging mostly at DEBUG but INFO also has a place.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/198)

### DIFF
--- a/src/churn/churn.ts
+++ b/src/churn/churn.ts
@@ -177,7 +177,7 @@ var log :logging.Log = new logging.Log('churn');
       return (random.randomUint32() % 255) + 1;
     } catch (e) {
       // https://github.com/uProxy/uproxy/issues/1593
-      log.debug('crypto unavailable, using Math.random');
+      log.warn('crypto unavailable, using Math.random');
       return Math.floor((Math.random() * 255)) + 1;
     }
   }
@@ -309,7 +309,7 @@ var log :logging.Log = new logging.Log('churn');
           this.peerName, endpoint);
       });
       this.onceHaveCaesarKey_.then((key: number) => {
-        log.debug('%1: caesar key is %2', this.peerName, key);
+        log.info('%1: caesar key is %2', this.peerName, key);
       });
     }
 
@@ -336,7 +336,6 @@ var log :logging.Log = new logging.Log('churn');
         remoteEndpoint:net.Endpoint,
         natEndpoints:NatPair,
         key:number) : void => {
-      log.debug('%1: configuring pipes...', this.peerName);
       this.pipe_ = freedom['churnPipe']();
       this.pipe_.setTransformer('caesar',
           new Uint8Array([key]).buffer,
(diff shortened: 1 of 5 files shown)


src/churn/churn.ts
```diff
@@ -351,9 +350,15 @@ var log :logging.Log = new logging.Log('churn');
       //       'ciphertext_dfa': regex2dfa('^.*$'),
       //       'ciphertext_max_len': 1450
       //     }
-      this.pipe_.bindLocal(natEndpoints.internal);
-      this.pipe_.setBrowserEndpoint(webRtcEndpoint);
-      this.pipe_.bindRemote(remoteEndpoint).then(this.haveForwardingSocketEndpoint_);
+      this.pipe_.bindLocal(natEndpoints.internal).then(() => {
```
bemasc: I'd rather not slow down this process by serializing it.  It's fine to log something at completion but let's keep the three Promises starting in parallel.

trevj: Good point! Done.

src/churn/churn.ts
```diff
@@ -351,9 +350,15 @@ var log :logging.Log = new logging.Log('churn');
       //       'ciphertext_dfa': regex2dfa('^.*$'),
       //       'ciphertext_max_len': 1450
       //     }
-      this.pipe_.bindLocal(natEndpoints.internal);
-      this.pipe_.setBrowserEndpoint(webRtcEndpoint);
-      this.pipe_.bindRemote(remoteEndpoint).then(this.haveForwardingSocketEndpoint_);
+      Promise.all<any>([
+          this.pipe_.bindLocal(natEndpoints.internal),
+          this.pipe_.setBrowserEndpoint(webRtcEndpoint),
+          this.pipe_.bindRemote(remoteEndpoint)]).then((answers:any[]) => {
```
bemasc: Technically, this could still be slower than the old implementation, because `this.haveForwardingSocketEndpoint_` now waits for all three promises, even though it only depends on one of them.  How about

```typescript
var bindRemote = this.pipe_.bindRemote(remoteEndpoint).then((mirrorEndpoint:net.Endpoint) : net.Endpoint => {
  this.haveForwardingSocketEndpoint_(mirrorEndpoint);
  return mirrorEndpoint;
});

Promise.all<any>([
    this.pipe_.bindLocal(natEndpoints.internal),
    this.pipe_.setBrowserEndpoint(webRtcEndpoint),
    bindRemote
])...
```

This would also avoid the need for a cast below.

trevj: I suggest we perform a separate micro-optimisation pass at some later time.

trevj: You convinced me :-p

bemasc:
:+1:

trevj:
Tiny change: log browser socket, remote endpoint, and mirror address separately. Now everything you would need to verify obufscation with Wireshark is in INFO logs. This was not the case in demo earlier today...

PR #200, opened 2015-06-12
Implement a per-user limit on the number of open sockets.

bemasc:
Fixes #1568

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/200)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-lib",
   "description": "Shared libraries for uProxy projects.",
-  "version": "27.2.5",
+  "version": "27.3.0",
   "repository": {
     "type": "git",
     "url": "https://github.com/uProxy/uproxy-lib"
--- a/src/integration-tests/socks-echo/base-spec.core-env.ts
+++ b/src/integration-tests/socks-echo/base-spec.core-env.ts
@@ -23,9 +23,9 @@ export function socksEchoTestDescription(useChurn:boolean) {
 
   var testerFactoryManager
         :freedom_types.FreedomModuleFactoryManager<ProxyIntegrationTester>;
-  var createTestModule = function(denyLocalhost?:boolean)
-      :ProxyIntegrationTester {
-    return testerFactoryManager(denyLocalhost, useChurn);
+  var createTestModule = function(denyLocalhost?:boolean,
+      sessionLimit?:number) : ProxyIntegrationTester {
+    return testerFactoryManager(denyLocalhost, useChurn, sessionLimit);
   };
 
   beforeEach((done) => {
@@ -67,12 +67,11 @@ export function socksEchoTestDescription(useChurn:boolean) {
       return testModule.echo(connectionId, input);
     }).then((output:ArrayBuffer) => {
       expect(arraybuffers.byteEquality(input, output)).toBe(true);
-      testModule.notifyClose(connId).then(() => {
-        testModule.on('sockClosed', (cnnid:string) => {
-          expect(cnnid).toBe(connId);
-          done();
-        });
-        testModule.closeEchoConnections();});
+      testModule.on('sockClosed', (cnnid:string) => {
+        expect(cnnid).toBe(connId);
+        done();
+      });
+      testModule.closeEchoConnections();
     });
   });
 
@@ -344,5 +343,24 @@ export function socksEchoTestDescription(useChurn:boolean) {
     }).catch((e:any) => {
       expect(e.reply).toEqual(socks.Reply.HOST_UNREACHABLE);
     }).then(done);
+  });
+
+  it('Hit the session limit', (done) => {
+    var limit = 4;
+    var testModule = createTestModule(false, limit);
+
+    testModule.startEchoServer().then((port:number) => {
+      for (var i = 0; i < limit; ++i) {
+        // Connections up to the limit are allowed.
+        testModule.connect(port).catch((e) => {
+          // These should not fail.
+          expect(e).toBeUndefined();
+        });
+      }
+      // This one is over the limit so it should fail to open.
+      testModule.connect(port).catch((e) => {
+        done();
+      });
+    });
   });
 };
--- a/src/integration-tests/socks-echo/freedom-module.json
+++ b/src/integration-tests/socks-echo/freedom-module.json
@@ -26,7 +26,7 @@
   "api": {
     "ProxyIntegrationTest": {
       "constructor": {
-        "value": ["boolean", "boolean"]
+        "value": ["boolean", "boolean", "number"]
       },
       "startEchoServer": {
         "type": "method",
@@ -62,10 +62,6 @@
           "connectionId": "string",
           "response": "buffer"
         }
-      },
-      "notifyClose": {
-        "type": "method",
-        "value": ["string"]
       },
       "closeEchoConnections": {
           "type": "method"
--- a/src/integration-tests/socks-echo/proxy-integration-test.ts
+++ b/src/integration-tests/socks-echo/proxy-integration-test.ts
@@ -76,7 +76,7 @@ class AbstractProxyIntegrationTest implements ProxyIntegrationTester {
     };
 
     if (typeof sessionLimit === 'number') {
-      (<any>(rtc_to_net.RtcToNet)).SESSION_LIMIT = sessionLimit;
+      rtc_to_net.RtcToNet.SESSION_LIMIT = sessionLimit;
     }
 
     this.socksToRtc_ = new socks_to_rtc.SocksToRtc();
--- a/src/integration-tests/socks-echo/proxy-integration-test.ts
+++ b/src/integration-tests/socks-echo/proxy-integration-test.ts
@@ -28,9 +28,11 @@ class AbstractProxyIntegrationTest implements ProxyIntegrationTester {
   private repeat_ :number = 1;
 
   constructor(private dispatchEvent_:(name:string, args:any) => void,
-                                      denyLocalhost?:boolean,
-                                      obfuscate?:boolean) {
-    this.socksEndpoint_ = this.startSocksPair_(denyLocalhost, obfuscate);
+              denyLocalhost?:boolean,
+              obfuscate?:boolean,
+              sessionLimit?:number) {
+    this.socksEndpoint_ = this.startSocksPair_(denyLocalhost, obfuscate,
+        sessionLimit);
   }
 
   public startEchoServer = () : Promise<number> => {
@@ -60,7 +62,8 @@ class AbstractProxyIntegrationTest implements ProxyIntegrationTester {
     return Promise.resolve<void>();
   };
 
-  private startSocksPair_ = (denyLocalhost?:boolean, obfuscate?:boolean) : Promise<net.Endpoint> => {
+  private startSocksPair_ = (denyLocalhost?:boolean, obfuscate?:boolean,
+      sessionLimit?:number) : Promise<net.Endpoint> => {
     var socksToRtcEndpoint :net.Endpoint = {
       address: this.localhost_,
       port: 0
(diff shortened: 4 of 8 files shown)


trevj:
Nice! A few things to consider but I like this!

---

Review status: 0 of 8 files reviewed, 8 unresolved discussions, some commit checks failed.

---

<sup>**[src/integration-tests/socks-echo/proxy-integration-test.ts, line 79 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdZi3I_gU7ljFujWUJ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/integration-tests/socks-echo/proxy-integration-test.ts#L79)):</sup>
I think it would be fine to make the limit public, even if it's just for testing purposes.

---

<sup>**[src/net/tcp.ts, line 392 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdZLnBRJ3Do8qxBjFH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/net/tcp.ts#L392)):</sup>
Why not just `onceClosed`?

---

<sup>**[src/net/tcp.ts, line 400 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdYpf_ABSJ5c2CjGeU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/net/tcp.ts#L400)):</sup>
Interesting. I've gone back and forth on whether interdependency of promises is a good idea. For example, something we've had for a while is that `onceClosed` fulfills if `onceOpened` rejects -- however, this complicates the code and is clearly unnecessary if the caller is willing to only set a `onceClosed` handler once `onceOpened` fulfills :-)

This case is less obviously redundant but still the interdependecy makes me a little uneasy! (this is fine though, please don't block based on my personal demons)

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 207 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-Jrd_W3xL9ewt55VPhMs)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L207)):</sup>
You don't need the `[]`.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 209 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdaZ2hpExKZpoNbmDU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L209)):</sup>
Interesting. How will `SocksToRtc` handle this? It would be nice if it returned one of the codes defined by the RFC, e.g. `X'05'` (connection refused):
https://tools.ietf.org/html/rfc1928#section-6

(of course any client trying to open 10k connections is either malicious or broken and we shouldn't go to too much trouble)

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 213 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-Jrd_davdeGVZOXNwP0m)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L213)):</sup>
How about combining `userSessionAdded_` with `userSessionAllowed` into a single function, e.g. `addUserSession`?

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 238 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-Jrd_9h369B7m_bHHhpm)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L238)):</sup>
If you supply the user ID as an argument then you can make this function static (which is usually good, especially when it essentially operates only on static data).

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 250 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-Jrd_FGaFGFS6DIBkI3W)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L250)):</sup>
Again, you can make this function static.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200)

bemasc:
Review status: 0 of 8 files reviewed at latest revision, 8 unresolved discussions, some commit checks pending.

---

<sup>**[src/integration-tests/socks-echo/proxy-integration-test.ts, line 79 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdZi3I_gU7ljFujWUJ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/integration-tests/socks-echo/proxy-integration-test.ts#L79)):</sup>
Done.

---

<sup>**[src/net/tcp.ts, line 392 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdZLnBRJ3Do8qxBjFH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/net/tcp.ts#L392)):</sup>
If we replaced `rejectOnClose` with `onceClosed`, then (a) the promise will fulfill with `undefined` instead of rejecting, and (b) we will lose any pending inbound data.  Maybe you have something else in mind?

---

<sup>**[src/net/tcp.ts, line 400 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdYpf_ABSJ5c2CjGeU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/net/tcp.ts#L400)):</sup>
FWIW, I made this change because otherwise it the integration test would just hang, rather than fail, if the socket closed while the client was waiting for a message.  I suppose I could have put this `Promise.race()` into the integration test instead, but having `receiveNext` return a `Promise` that might never fulfill or reject seemed weird to me.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 207 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-Jrd_W3xL9ewt55VPhMs)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L207)):</sup>
Done.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 209 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdaZ2hpExKZpoNbmDU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L209)):</sup>
`SocksToRtc` closes the TCP socket in response, without reporting any error code.  I'm not sure whether reporting an error code would be helpful.  Arguably, closing the socket without an error code is more "expressive", allowing us to distinguish different kinds of failures.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 213 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-Jrd_davdeGVZOXNwP0m)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L213)):</sup>
Done.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 238 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-Jrd_9h369B7m_bHHhpm)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L238)):</sup>
Done.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 250 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-Jrd_FGaFGFS6DIBkI3W)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L250)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200)

trevj:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

<sup>**[src/net/tcp.ts, line 392 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdZLnBRJ3Do8qxBjFH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/net/tcp.ts#L392)):</sup>
Ah, yes, I forgot `onceClosed` never rejects.

Don't you *always* want to reject when the socket closes, though? Right now it looks like `rejectOnClose` itself won't always fulfill or reject.

---

<sup>**[src/net/tcp.ts, line 400 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdYpf_ABSJ5c2CjGeU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/net/tcp.ts#L400)):</sup>
You've reminded me to go read up on best practices for asynchronous programming.

This is fine but it seems to me like we're slowly weaving a tangled web of tightly interconnected promises here.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 240 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdaZ2hpExKZpoNbmDU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/rtc-to-net/rtc-to-net.ts#L240)):</sup>
Well, it's another way to express what happened but probably not ideal!

Since this is just protection against rogue clients though this seems fine.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200)

bemasc:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[src/net/tcp.ts, line 392 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdZLnBRJ3Do8qxBjFH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/net/tcp.ts#L392)):</sup>
`Promise.race` ensures that the ultimate return value always succeeds or rejects ... but never mind.  This code is clearly hard to understand, so I've simplified it.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200)

trevj:
:+1:

---

Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/net/tcp.ts, line 390 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200#-JrdZLnBRJ3Do8qxBjFH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/890aeaad68dc1d19bdb79715215b4b950ea32baa/src/net/tcp.ts#L390)):</sup>
OK, thanks for explaining in person!

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/200)

bemasc:
Nope.  Now uProxy needs to pass the userId to the RtcToNet constructor.

PR #204, opened 2015-06-18
Use an ISO-8601 string for log timestamp

jpevarnek:
This simplifies the process of getting a timestamp for the log as well
as switching to using UTC times.

Fixes uProxy/uproxy#1650

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/204)


trevj:
Thanks for looking at this!

So, let's update the comments too, e.g.:
https://github.com/uProxy/uproxy-lib/blob/iso-for-timestamp/src/loggingprovider/loggingprovider.ts#L123

jpevarnek:
PTAL

trevj:
:+1:

PR #211, opened 2015-06-29
remove (SocksToRtc|RtcToNet).startFromConfig

trevj:
In the interest of having smaller interfaces, and how the args for `SocksToRtc` are threatening to spiral out of control as we add more and more options...remove this.

uProxy doesn't use these functions so it's just fine.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/211)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-lib",
   "description": "Shared libraries for uProxy projects.",
-  "version": "27.5.0",
+  "version": "28.0.0",
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


bemasc:
:+1:
Do remember to remove `startFromConfig` from the mocks in uProxy/uproxy too.

trevj:
Good point, will do!

trevj:
I've prepared a pull request:
https://github.com/uProxy/uproxy/compare/trevj-uproxy-lib-v28?expand=1

PR #212, opened 2015-06-29
Only bind to the ANY interface, because this is all Firefox supports anyway

bemasc:
This unblocks https://github.com/freedomjs/freedom-for-firefox/pull/68, which would otherwise break this code.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/212)

### DIFF
--- a/src/churn-pipe/churn-pipe.ts
+++ b/src/churn-pipe/churn-pipe.ts
@@ -172,11 +172,17 @@ class Pipe {
     log.debug('Binding public endpoint: %1', publicEndpoint);
     var socket = freedom['core.udpsocket']();
     var index = this.addPublicSocket_(socket, publicEndpoint);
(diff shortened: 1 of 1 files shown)


src/churn-pipe/churn-pipe.ts
```diff
@@ -290,6 +296,10 @@ class Pipe {
     return Promise.all(promises).then((fulfills:any[]) : void => {});
   }
 
+  private static anyInterface_ = (address:string) => {
```
trevj: maybe a comment here?

trevj:
:+1:

src/churn-pipe/churn-pipe.ts
bemasc: Done.

PR #213, opened 2015-06-30
Remove reliance on nonexistent UDP binding return code.

bemasc:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/213)


trevj:
Good find! `bind` does at least throw an exception if an error occurs? :$

:+1:

bemasc:
`bind` is supposed to reject if there's a problem, yes.  This is already true in Chrome, and soon will be in Firefox too: https://github.com/freedomjs/freedom-for-firefox/pull/71

PR #215, opened 2015-06-30
make sample apps listen on all ports and slow down the control port a little

trevj:
I'm finding 500ms is a little too short to reliably gather candidates...this seems to work out better. Until we can re-write the copy/paste app a little bit :-)

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/215)


bemasc:
:+1:

PR #216, opened 2015-07-01
Have proxy listen to INADDR_ANY instead of localhost

mollyling:
For docker support.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/216)


trevj:
:+1:

PR #219, opened 2015-07-08
SOCKS adventure, for automated testing

trevj:
This adds SOCKS Adventure, intended for automated testing. Essentially, it factors out the control port stuff from copypaste into its own app. Tested with Chrome on Firefox.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/219)


trevj:
I forgot to mention that I can remove the current control port stuff from `copypaste` in a followup (let's not break the current docker stuff).

bin/connect-pair.py
```diff
@@ -1,21 +1,38 @@
...
-first.send(second_sdp)
+getter.sendall('get\n')
+giver.sendall('give\n')
+
+getterOpen = True
+giverOpen = True
+while getterOpen and giverOpen:
+  inready, outready, exceptready = select.select([getter, giver], [], [])
+  for s in inready:
+    raw = s.recv(4096)
```
mollyling: Perhaps make the sockets nonblocking (https://docs.python.org/2/howto/sockets.html#non-blocking-sockets) and `read()` until you don't get data back (or just < 4k).  Otherwise you may splice the last line of a large block with a newline below.

src/adventure/freedom-module.ts
```diff
@@ -0,0 +1,184 @@
...
+        break;
+      case 'give':
+        give(connection);
+        break;
+      case 'ping':
+        sendReply('ping', connection);
+        connection.dataFromSocketQueue.setSyncNextHandler(processCommands);
+        break;
+      case 'xyzzy':
+        sendReply('Nothing happens.', connection);
```
mollyling: :-)

bin/connect-pair.py
trevj: Are you worried signalling channel messages  >4096 bytes? I don't think that will ever happen. What if I just read more bytes, e.g. 8192?

trevj: As discussed in person, let's make it 16K and see if we encounter any weirdness :-)

trevj:
Added a note to the README. PTAL!

trevj:
Got a verbal thumbs up. Thanks!

PR #223, opened 2015-07-14
add bower dependency, so that it does not have to be installed globally

trevj:
Oopsy, turns out we actually still do need bower installed globally.

I stole this fix from uProxy:
https://github.com/uProxy/uproxy/blob/dev/setup.sh#L7

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/223)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   "devDependencies": {
     "arraybuffer-slice": "^0.1.2",
     "babar": "^0.0.3",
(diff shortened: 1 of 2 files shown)


trevj:
Tested on my Mac box as well as on my new Windows virtual machine, where I noticed this.

trevj:
Got a verbal +1 from @mollyling .

PR #228, opened 2015-07-22
Version number 28.0.0 => 29.0.0

kennysong:
Bumped the version number – uproxy-lib now includes the freedom-port-control module.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/228)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-lib",
   "description": "Shared libraries for uProxy projects.",
-  "version": "28.0.0",
+  "version": "29.0.0",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


bemasc:
:+1:

trevj:
Not a big deal at all but FYI when adding new APIs you only need to bump the *minor*, i.e. middle, number -- only *breaking* changes require a major version bump (and the ensuing changes to uproxy's `package.json`).

bemasc:
This arguably required a major version bump because it would break uproxy at runtime until uproxy changed its freedom manifest.

trevj:
OK. It's not obvious why that would be the case but thanks for clarifying.

PR #230, opened 2015-07-24
Added packet length transformers and changed Transformer API to use lists of ArrayBuffers.

blanu:
This is just a prototype. I have not written any unit tests. I have no checked for conformance to the style guide. I have not refactored the code to remove duplicated methods. I'm sending this patch in early to get some feedback and direction before I start cleaning up the code.

I added several transformers to act as functional tests. I have the following functionality in the transformers:
- add and remove fragmentation headers, without otherwise modifying the packet
- always increase the size, adding padding
- always decrease the size (causing fragmentation)
- normalize all packets to the same size
- randomize packet size within a specified range using a uniform distribution
- randomize packet size using a multinomial distribution

Additionally, the Transformer interface has been changed so that transform and restore now return ArrayBuffer[] instead of ArrayBuffer. This is to allow for fragmentation and defragmentation.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/230)


src/churn-pipe/churn-pipe.ts
```diff
@@ -60,6 +69,18 @@ var makeTransformer_ = (
      transformer = Fte.Transformer();
      } else */ if (name == 'caesar') {
```
mollyling: I feel that we could turn this into a map from string to constructor.

src/churn/churn.ts
```diff
@@ -271,9 +273,34 @@ export var filterCandidatesFromSdp = (sdp:string) : string => {
     private configurePipe_ = (key:number) : void => {
       this.pipe_ = freedom['churnPipe'](this.peerName);
       this.pipe_.on('mappingAdded', this.onMappingAdded_);
-      this.pipe_.setTransformer('caesar',
+      /*this.pipe_.setTransformer('none',
```
mollyling: This is all a bit messy.  There seems to be quite a bit of magic going on in here.

src/churn/churn.ts
```diff
@@ -288,6 +315,17 @@ export var filterCandidatesFromSdp = (sdp:string) : string => {
       this.havePipe_();
     }
 
+    private makeSampleTargetLengthDistribution_ = () : Array<number> => {
+      var result : Array<number> =[];
```
mollyling: Style guide: types have colons right next to them.  So ":Array<number>" not ": Array<number>."  Not that you were claiming to adhere to the guide, but FYI.

src/fancy-transformers/defragmenter.ts
```diff
@@ -0,0 +1,87 @@
+import logging = require('../logging/logging');
+import arraybuffers = require('../arraybuffers/arraybuffers');
+import Fragment = require('./fragment');
+
+var log :logging.Log = new logging.Log('fancy-transformers');
+
+class Defragmenter {
```
mollyling: A class description would be nice.

src/fancy-transformers/defragmenter.ts
```diff
@@ -0,0 +1,87 @@
...
+  private tracker_ : {[index: string]: ArrayBuffer[]}={};
+  private counter_ : {[index: string]: number}={};
+  private complete_: string[]=[];
```
mollyling: Again, style guide for spacing around colons.

src/churn/churn.ts
```diff
@@ -288,6 +315,17 @@ export var filterCandidatesFromSdp = (sdp:string) : string => {
...
+      var prob = 1 / 1440;
+      var index=0;
```
mollyling: There should be spaces around operators.

src/fancy-transformers/defragmenter.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  public constructor() {}
+
+  public addFragment(fragment:Fragment) {
+    var hexid=arraybuffers.arrayBufferToHexString(fragment.id);
+    if(hexid in this.tracker_) { // A fragment for an existing packet
```
mollyling: Vertical space is cheap, visual bandwidth is expensive.  Please put the comments on their own lines.

src/fancy-transformers/fragment.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+var split_ = (buffer:ArrayBuffer, firstLen:number) : Array<ArrayBuffer> => {
+  var bytes=new Uint8Array(buffer)
+  var lastLen : number = buffer.byteLength-firstLen;
+  var first = new Uint8Array(firstLen);
+  var last = new Uint8Array(lastLen);
+  var fromIndex : number = 0;
+  var toIndex : number = 0;
+  while(toIndex < first.length) {
```
mollyling: Why isn't this a for() loop?

src/fancy-transformers/fragment.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  return [first.buffer, last.buffer];
+}
+
+/* Takes a two byte (network byte order) representation of a number and returns
+ * the number.
+ */
+ // TODO(bwiley): Byte order may be backward
+ // TODO(bwiley): Fix type error
+var decodeLength_ = (buffer:ArrayBuffer) : number => {
```
mollyling: Perhaps generalize to ntohs/htons()?  `ntohs(buffer:ArrayBuffer, offset:int) :number` and `htons(value:number, buffer:ArrayBuffer, offset:int) :void`?

src/fancy-transformers/fragment.ts
```diff
@@ -0,0 +1,177 @@
...
+      bytes[i] = Math.floor(Math.random()*255);
+    }
+    return bytes.buffer;
+  }
+
+  static decodeFragment = (buffer:ArrayBuffer) : Fragment => {
+//    log.debug('Decode fragment %1 %2', buffer.byteLength, length);
+    var parts = split_(buffer, 2);
+    var lengthBytes = parts[0];
+    var length = decodeLength_(lengthBytes);
```
mollyling: These two variables: length and lengthBytes, don't tell me much about what they're the lengths of.

src/fancy-transformers/fragment.ts
```diff
@@ -0,0 +1,177 @@
...
+      total=total+buffers[i].byteLength;
+    }
+
+    var result = new Uint8Array(total);
+    var toIndex=0;
+    for(var i=0; i<buffers.length; i++) {
+      var bytes=new Uint8Array(buffers[i]);
+      for(var fromIndex=0; fromIndex<buffers[i].byteLength; fromIndex++) {
+        result[toIndex]=bytes[fromIndex];
+        toIndex=toIndex+1;
```
mollyling: you can do fromIndex++, toIndex++ in the last clause of the for-loop header.  Also another good place for a general copy routine.

src/fancy-transformers/defragmenter.ts
```diff
@@ -0,0 +1,87 @@
...
+        fragments[fragment.index]=fragment.payload;
+        this.tracker_[hexid]=fragments;
+        this.counter_[hexid]=this.counter_[hexid]-1;
+        if(this.counter_[hexid]==0) {
+          this.complete_.push(hexid);
+        }
+      }
+    } else { // A fragment for a new packet
+      var fragments : ArrayBuffer[]=[];
+      for(var i=0; i<fragment.count; i++) {
```
mollyling: I donno if you care about the difference but Array has a constructor that takes a single numerical arg, and results in an array of size `arg`, each initialized to `undefined`.

src/fancy-transformers/defragmenter.ts
```diff
@@ -0,0 +1,87 @@
...
+        fragments.push(null);
+      }
+
+      fragments[fragment.index]=fragment.payload;
+      this.tracker_[hexid]=fragments;
+      this.counter_[hexid]=fragment.count-1;
+      if(this.counter_[hexid]==0) {
+        this.complete_.push(hexid);
```
mollyling: Is this array intended to be grow-only, for the lifetime of the connection?

blanu:
This pull request is out of date. I am closing it and creating an up-to-date one.

PR #222, opened 2015-07-10
Added port control module to uproxy-lib

kennysong:
We now try to use NAT-PMP, PCP, and UPnP to open ports when making a connection inside churn.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/222)

PR #234, opened 2015-07-28
fix adventure app by adding port control

trevj:
Followup to:
https://github.com/uProxy/uproxy-lib/pull/222

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/234)


kennysong:
👍

PR #235, opened 2015-07-31
echo server cleanup

trevj:
Just a bit of noodling around -- the sample apps should be as simple as possible, IMHO.

Also, add some logging for when a client disconnects. This wasn't here, invalidating a bunch of testing I'd done against freedom-for-firefox. Oops.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/235)


trevj:
@bemasc, maybe you can take a look...super low priority, just noodling about in the code garden.

bemasc:
:+1:

PR #240, opened 2015-08-05
move the NAT diagnosis code from uproxy-probe into this repo

trevj:
So, I needed a break from Docker stuff and finally made a start on this:
https://github.com/uProxy/uproxy/issues/1509

Having combed through the code, I just copied the `doNatProvoking()` function -- `doUdpTest` and  `doStunAccessTest` are super-low-level functions we can do with standard tools (`ping` and `stunclient` respectively). I renamed `doNatProvoking` -> `probe`.

Also, a lightweight Chrome and Firefox app called "uprobe" (console output only, I figure we don't use this often enough to justify any much more).

Tested with uProxy itself; pull request to follow.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/240)


trevj:
I almost forgot: not sure yet what to do with the *server* code in uproxy-probe. No reason it couldn't move in here too, at some point.

src/nat/probe.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    var rejectShortcut: (e: any) => void = null;
+
+    function onUdpData(info: freedom_UdpSocket.RecvFromInfo) {
+      var rspStr: string = arraybuffers.arrayBufferToString(info.data);
+      log.debug('receive response = ' + rspStr);
+
+      var rsp = JSON.parse(rspStr);
+
+      if (rsp['answer'] == 'FullCone') {
```
bemasc: I think the style guide demands ===.

bemasc:
:+1: but maybe we should just have a general issue to do a style cleanup.

src/nat/probe.ts
trevj: Yep, fixed. As discussed, I've filed an issue to perform a deeper cleanup:
https://github.com/uProxy/uproxy/issues/1808

PR #241, opened 2015-08-06
Tolerate duplicate negotiationneeded events in Firefox 40

bemasc:
This is half of a fix to https://github.com/uProxy/uproxy/issues/1777 .  (The other half is change to freedom-for-firefox.)

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/241)


trevj:
I feel we are being to prescriptive with our checking of state but I tested this on my system and it seems to work great!

:+1:

PR #243, opened 2015-08-11
uproxy-lib build script for windows

kennysong:
`setup.cmd` will do the same thing as `setup.sh`, but on Windows. 

The dependency `tsd` has also been updated from `0.5.7` to `0.6.3`, since the new version removes an (optional) dependency on Python 2.7, which would error on vanilla Windows.

The readme has also been updated.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/243)


kennysong:
@trevj Added a new change to the build script – apparently it went into this branch again.

trevj:
:+1:

Thanks!

trevj:
Oops, never merged this in. To be honest I'm sceptical we'll keep this up to date over time but since Windows is popular and we haven't changed the build system for a while...let's try it for a bit.

PR #244, opened 2015-08-11
UPnP failures will be logged now

kennysong:
We now `log.debug` a UPnP failure mapping object, so it's included in the submit feedback logs.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/244)

### DIFF
--- a/src/churn/churn.ts
+++ b/src/churn/churn.ts
@@ -231,8 +231,8 @@ export var filterCandidatesFromSdp = (sdp:string) : string => {
               this.portControl_.addMapping(c.relatedPort, c.port, MAP_LIFETIME).
                   then((mapping:freedom_PortControl.Mapping) => {
                     if (mapping.externalPort === -1) {
-                      log.debug("addMapping() failed.");
-                      log.debug("Failed UPnP mapping object: ", mapping);
+                      log.debug("addMapping() failed. UPnP mapping object: ", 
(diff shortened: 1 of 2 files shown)


src/churn/churn.ts
```diff
@@ -232,6 +232,7 @@ export var filterCandidatesFromSdp = (sdp:string) : string => {
                   then((mapping:freedom_PortControl.Mapping) => {
                     if (mapping.externalPort === -1) {
                       log.debug("addMapping() failed.");
+                      log.debug("Failed UPnP mapping object: ", mapping);
```
bemasc: This should just be one log line, not two.

kennysong:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[src/churn/churn.ts, line 235 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/244#-JwThP81jPKfQ-alxS4l-r1-235)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/b6c39c8ca769d367949269a479aabb3f1c076ccf/src/churn/churn.ts#L235)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/244)

PR #245, opened 2015-08-14
Churn uses uproxy core's port control module

kennysong:
Instead of instantiating its own version of `freedom-port-control`, churn now takes in the module as an optional argument in the constructor. This way, we can use only one instance of the module, which is instantiated and stored inside uproxy core.

Everything on the `uproxy-lib` side of this change should be done.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/245)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-lib",
   "description": "Shared libraries for uProxy projects.",
-  "version": "29.0.6",
+  "version": "29.0.7",
   "repository": {
     "type": "git",
     "url": "https://github.com/uProxy/uproxy-lib"
--- a/src/bridge/bridge.ts
+++ b/src/bridge/bridge.ts
@@ -1,5 +1,6 @@
 /// <reference path='../../../third_party/typings/es6-promise/es6-promise.d.ts' />
 /// <reference path='../../../third_party/freedom-typings/freedom-common.d.ts' />
+/// <reference path='../../../third_party/freedom-typings/port-control.d.ts' />
 
 import churn = require('../churn/churn');
 import churn_types = require('../churn/churn.types');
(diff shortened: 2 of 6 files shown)


bemasc:
:+1: 
Also need to change copypaste-socks, and bump the minor version in package.json.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/245)

src/bridge/bridge.ts
```diff
@@ -208,7 +215,7 @@ export class BridgingPeerConnection implements peerconnection.PeerConnection<
       pc:freedom_RTCPeerConnection.RTCPeerConnection)
       :peerconnection.PeerConnection<churn_types.ChurnSignallingMessage> => {
     log.debug('%1: constructing churn peerconnection', this.name_);
-    return new churn.Connection(pc, this.name_);
+    return new churn.Connection(pc, this.name_, undefined, this.portControl_);
```
trevj: I think `false` is more appropriate for the third argument.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-lib",
   "description": "Shared libraries for uProxy projects.",
-  "version": "29.0.6",
+  "version": "29.0.7",
```
trevj: Since this changes the behaviour for existing clients, I think a major version bump is more appropriate.

src/bridge/bridge.ts
```diff
@@ -18,26 +19,31 @@ var log :logging.Log = new logging.Log('bridge');
 // you don't want to use obfuscation.
 export var preObfuscation = (
     name?:string,
-    config?:freedom_RTCPeerConnection.RTCConfiguration)
+    config?:freedom_RTCPeerConnection.RTCConfiguration,
+    portControl?:freedom_PortControl.PortControl)
```
trevj: This is a little weird since only churn makes use of port control, right?

PR #248, opened 2015-08-17
Added fancy transformers and changed the Transformer API

blanu:
This branch includes transformers for packet length (with fragmentation), byte sequence injection, encryption, and compression.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/248)


bemasc:
Could you separate the whitespace changes into a separate pull request?

Also, the CI is failing, which seems like a bad sign.

blanu:
The CI seems to be failing for apparently unrelated reasons. The CI seems to be broken. "Storage server temporarily offline. See https://status.github.com for GitHub system status."

blanu:
Regarding whitespace changes, these were requested in response to my previous pull request, which is why they are included in this pull request.

iislucas:
Cool! I should probably do my review on the real pull request instead of this one. I reviewed just the array buffers changes on this one. Send me the new PRs and I'll review them. 

---

Reviewed 2 of 24 files at r3.
Review status: 2 of 25 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

<sup>**[src/arraybuffers/arraybuffers.ts, line 122 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/248#-Jx1GTVssisp2ysj5__B)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/986f63bd4a8f5a602f2bb3f3db04563a2cf824b0/src/arraybuffers/arraybuffers.ts#L122)):</sup>
The name of this function is a little abstract, would be good to have a comment to say what this is doing. e.g. how is it different to concat. 

---

<sup>**[src/arraybuffers/arraybuffers.ts, line 141 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/248#-Jx1H9VcgWIdXvH4gKHr)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/986f63bd4a8f5a602f2bb3f3db04563a2cf824b0/src/arraybuffers/arraybuffers.ts#L141)):</sup>
Don't return an array when you want it to be a pair: better to use a real interface/structure. but note: For array's use the "Type[]" style. 

---

<sup>**[src/arraybuffers/arraybuffers.ts, line 171 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/248#-Jx1HWAcKhxhYqQyChSX)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/986f63bd4a8f5a602f2bb3f3db04563a2cf824b0/src/arraybuffers/arraybuffers.ts#L171)):</sup>
There's a real function for doing this in crypto. Do we really need to have this (not crypto) implementation also?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/248)

PR #250, opened 2015-08-18
Transformer returns multiple ArrayBuffers from transform and restore

blanu:
The Transformer API has been changed and all of the simple-transformers have been updated to use the new API. churn-pipe has also been updated to handle transformers that return multiple ArrayBuffers.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/250)

### DIFF
--- a/src/churn-pipe/churn-pipe.ts
+++ b/src/churn-pipe/churn-pipe.ts
@@ -211,7 +211,7 @@ class Pipe {
         this.onIncomingData_(recvFromInfo, publicEndpoint.address, index);
       });
     });
-    
+
     this.publicPorts_[publicEndpoint.address][publicEndpoint.port] = portPromise;
     return portPromise;
   }
@@ -411,12 +411,14 @@ class Pipe {
    * The message is obfuscated before it hits the wire.
    */
   private sendTo_ = (publicSocket:Socket, buffer:ArrayBuffer, to:net.Endpoint)
-      : void => {
-    var transformedBuffer = this.transformer_.transform(buffer);
-    publicSocket.sendTo.reckless(
-      transformedBuffer,
-      to.address,
-      to.port);
+      :void => {
+    var transformedBuffers = this.transformer_.transform(buffer);
+    for(var i=0; i<transformedBuffers.length; i++) {
+      publicSocket.sendTo.reckless(
+        transformedBuffers[i],
+        to.address,
+        to.port);
+    }
   }
 
   /**
@@ -432,12 +434,14 @@ class Pipe {
       return;
     }
     var transformedBuffer = recvFromInfo.data;
-    var buffer = this.transformer_.restore(transformedBuffer);
+    var buffers = this.transformer_.restore(transformedBuffer);
     this.getMirrorSocket_(recvFromInfo, index).then((mirrorSocket:Socket) => {
-      mirrorSocket.sendTo.reckless(
-          buffer,
-          iface,
-          browserPort);
+      for(var i=0; i<buffers.length; i++) {
+        mirrorSocket.sendTo.reckless(
+            buffers[i],
+            iface,
+            browserPort);
+      }
     });
   }
 
--- a/src/churn-pipe/churn-pipe.ts
+++ b/src/churn-pipe/churn-pipe.ts
@@ -413,7 +413,7 @@ class Pipe {
   private sendTo_ = (publicSocket:Socket, buffer:ArrayBuffer, to:net.Endpoint)
       :void => {
     var transformedBuffers = this.transformer_.transform(buffer);
-    for(var i=0; i<transformedBuffers.length; i++) {
+    for(var i = 0; i < transformedBuffers.length; i++) {
       publicSocket.sendTo.reckless(
         transformedBuffers[i],
         to.address,
@@ -436,7 +436,7 @@ class Pipe {
     var transformedBuffer = recvFromInfo.data;
     var buffers = this.transformer_.restore(transformedBuffer);
     this.getMirrorSocket_(recvFromInfo, index).then((mirrorSocket:Socket) => {
-      for(var i=0; i<buffers.length; i++) {
+      for(var i = 0; i < buffers.length; i++) {
         mirrorSocket.sendTo.reckless(
             buffers[i],
             iface,
--- a/src/simple-transformers/caesar.spec.ts
+++ b/src/simple-transformers/caesar.spec.ts
@@ -37,7 +37,7 @@ describe("caesar cipher", function() {
   it('transform buffer', function() {
     setKey(1);
     var bytes = new Uint8Array([4, 1, 255]);
-    var result = new Uint8Array(transformer.transform(bytes.buffer));
+    var result = new Uint8Array(transformer.transform(bytes.buffer)[0]);
     // toEquals() doesn't work for Uint8Array.
     expect(result[0]).toEqual(5);
     expect(result[1]).toEqual(2);
@@ -49,7 +49,7 @@ describe("caesar cipher", function() {
 
     setKey(1);
     var output = new Uint8Array(transformer.restore(
-        transformer.transform(new Uint8Array([4, 1, 255]).buffer)));
+        transformer.transform(new Uint8Array([4, 1, 255]).buffer)[0])[0]);
 
     for (var i = 0; i < output.length; i++) {
       expect(output[i]).toEqual(input[i]);
--- a/src/simple-transformers/caesar.ts
+++ b/src/simple-transformers/caesar.ts
@@ -34,14 +34,14 @@ class CaesarCipher implements Transformer {
   /** Nothing to configure -- all this transformer needs is a key. */
   public configure = (json:string) : void => {}
 
-  public transform = (buffer:ArrayBuffer) : ArrayBuffer => {
+  public transform = (buffer:ArrayBuffer) : ArrayBuffer[] => {
     this.map_(buffer, this.transformByte);
-    return buffer;
+    return [buffer];
   }
 
-  public restore = (buffer:ArrayBuffer) : ArrayBuffer => {
+  public restore = (buffer:ArrayBuffer) : ArrayBuffer[] => {
     this.map_(buffer, this.restoreByte);
-    return buffer;
+    return [buffer];
   }
 
   // No-op (we have no state or any resources to dispose).
--- a/src/simple-transformers/passthrough.ts
+++ b/src/simple-transformers/passthrough.ts
@@ -6,16 +6,18 @@
 /** An obfuscator which does nothing. */
 class PassThrough implements Transformer {
 
+  public constructor() {}
+
   public setKey = (key:ArrayBuffer) => {}
 
   public configure = (json:string) : void => {}
 
-  public transform = (buffer:ArrayBuffer) : ArrayBuffer => {
-    return buffer;
+  public transform = (buffer:ArrayBuffer) : ArrayBuffer[] => {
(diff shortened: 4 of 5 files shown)


trevj:
Awesome, thanks for splitting this into smaller pieces.

I have a suggestion for how an async interface will look but as-is I think we could proceed with this.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[src/simple-transformers/passthrough.ts, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1QIU_HU3pp9nkZfot)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/6e44d1324d4495b8d86be1864d33be1eec072f86/src/simple-transformers/passthrough.ts#L9)):</sup>
Did you find you needed this?

---

<sup>**[third_party/uTransformers/utransformers.d.ts, line 34 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1QQKvN40Mz3n0yfCQ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/6e44d1324d4495b8d86be1864d33be1eec072f86/third_party/uTransformers/utransformers.d.ts#L34)):</sup>
There's something unsatisfying about this interface. If you reckon this will support your plans then I'm okay with it -- however, if we're going to take the trouble to change this interface then I wonder if it's worth going the whole hog and implementing the async interface we briefly discussed. Up to you. We can make the interface pretty later. I don't want to block much longer on submitting this!

If not, you should mention how the returned array will be empty if - for example due to fragmentation - no data is to be transformed at this time. Perhaps an empty array indicates "no data at this time" while `null` indicates failure?

FWIW, I think the async interface will look like this:

```typescript
interface Transformer {
  // ... setKey, boring stuff, etc.

  // Queues some data for obfuscation.
  transform(buffer:ArrayBuffer) : void;

  // Sets a function to be invoked when obfuscated data is ready to be sent.
  setTransformedCallback(f:(buffer:ArrayBuffer) => void) : void;

  // Queues some data for de-obfuscation.
  restore(buffer:ArrayBuffer) : void;

  // Sets a function to be invoked when de-obfuscated is ready to be sent.
  setRestoredCallback(f:(buffer:ArrayBuffer) => void) : void;
}
```

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250)

blanu:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[src/simple-transformers/passthrough.ts, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1QIU_HU3pp9nkZfot)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/6e44d1324d4495b8d86be1864d33be1eec072f86/src/simple-transformers/passthrough.ts#L9)):</sup>
Yes, this is required for a change that is in a different pull request. I refactored the big if-else block of transformer initializers in churn-pipe into a map from names to constructor functions. This requires that all transformers have a constructor function in order to typecheck.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250)

trevj:
OK, I guess you're opting against the async change right now.

One or two tiny comments/suggestions, then :+1:.

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

<sup>**[src/churn-pipe/churn-pipe.ts, line 416 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1lZOH7YoYv64azKAY)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ed7ba456dbdbd21e654ae4779fc28b7f59d9fab2/src/churn-pipe/churn-pipe.ts#L416)):</sup>
style:
```typescript
for (var i = 0; i < transformedBuffers.length; i++) {
```

---

<sup>**[src/churn-pipe/churn-pipe.ts, line 439 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1ldh3v6am4FhPOR5X)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ed7ba456dbdbd21e654ae4779fc28b7f59d9fab2/src/churn-pipe/churn-pipe.ts#L439)):</sup>
same style comment as above

---

<sup>**[third_party/uTransformers/utransformers.d.ts, line 24 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1mNUCXCDi7aNOWMcJ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ed7ba456dbdbd21e654ae4779fc28b7f59d9fab2/third_party/uTransformers/utransformers.d.ts#L24)):</sup>
How is an error signaled? Via an exception?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250)

iislucas:
Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[src/simple-transformers/passthrough.ts, line 9 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1sMG_nZRUoO3PNKIO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ed7ba456dbdbd21e654ae4779fc28b7f59d9fab2/src/simple-transformers/passthrough.ts#L9)):</sup>
I don't think an explicit empty constructor adds anything... am I wrong? I'd like to know if I'm wrong! :) 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250)

blanu:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[src/simple-transformers/passthrough.ts, line 9 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1sMG_nZRUoO3PNKIO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ed7ba456dbdbd21e654ae4779fc28b7f59d9fab2/src/simple-transformers/passthrough.ts#L9)):</sup>
This is required for typechecking to succeed on a future change. When we construct the map from names to transformer constructor functions in churn-pipe, PassThrough won't typecheck unless it has a constructor function.

---

<sup>**[third_party/uTransformers/utransformers.d.ts, line 24 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1mNUCXCDi7aNOWMcJ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ed7ba456dbdbd21e654ae4779fc28b7f59d9fab2/third_party/uTransformers/utransformers.d.ts#L24)):</sup>
Great question! There are several sub-questions:
How did the code behave before I changed it, according to the documentation?

Errors were signaled by an empty ArrayBuffer.

How did the code behave before I changed it, in reality?

The caller does no error checking for this condition and the simple transformers never return an error.
So essentially an empty ArrayBuffer is treated like any other ArrayBuffer.

How does the code behave now, according to the documentation?

There is no explicit error signaling. The transformer may return an empty array.

How did the code behave now, in reality?

The caller does no error checking and the simple transformers never return an error.
So essentially an empty array is treated like any other array.

How should the code behave ideally?

The uProxy style guide specifies not to use null return values or throw exceptions and instead to include the possibility of errors in the return type of the function.
Determining this type requires some thought as to what the ideal failure mode is for transformers. Currently according to the type signature, transformers cannot
explicitly fail. They can only return no data. Subsequent calls may return data. So transient errors are hidden from the caller. Permanent failures result in the
transformer returning no data permanently. An alternate failure mode would be to have the transformer explicitly signal the caller of failure. In this case, how
should the caller respond? Should it re-initialize the transformer? Should it raise a UI alert to the user?

In summary, there is much to discuss about the ideal Transformer API. As far as this pull request goes, it brings the documentation up to sync with what was actually
implemented, which is that there is no explicit error signaling and errors manifest as empty array return values.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250)

blanu:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250)

trevj:
Review status: 4 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[third_party/uTransformers/utransformers.d.ts, line 24 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1mNUCXCDi7aNOWMcJ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ed7ba456dbdbd21e654ae4779fc28b7f59d9fab2/third_party/uTransformers/utransformers.d.ts#L24)):</sup>
What a mess. That style guide entry is...interesting.

I *think* the "null for error" was for uTransformers' benefit, where it isn't possible to propagate an exception from emscripten-ed code...although skimming the JS wrapper now, I actually don't see why the wrapper couldn't raise an exception:
https://github.com/uProxy/uproxy-obfuscators/blob/master/src/transformers/transformer.js

I personally hate "null for error" and it looks like `caesar.ts` doesn't implement it anyway. OK, let's just not mention exceptions for now, leaving the possibility of exceptions - as always - implicit.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250)

iislucas:
:+1: 

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250)

iislucas:
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

<sup>**[third_party/uTransformers/utransformers.d.ts, line 24 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1mNUCXCDi7aNOWMcJ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ed7ba456dbdbd21e654ae4779fc28b7f59d9fab2/third_party/uTransformers/utransformers.d.ts#L24)):</sup>
I think I wrote that entry in the style guide. I think it's sensible, but not as well written as it could be. Happy to discuss. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250)

trevj:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[third_party/uTransformers/utransformers.d.ts, line 24 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250#-Jx1mNUCXCDi7aNOWMcJ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ed7ba456dbdbd21e654ae4779fc28b7f59d9fab2/third_party/uTransformers/utransformers.d.ts#L24)):</sup>
Definitely one to discuss! FWIW, I agree exceptions for control flow are often bad and that JavaScript's exception mechanism is kinda crappy...

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/250)

PR #252, opened 2015-08-19
use an explicit channel ID to theoretically double the number of channels

trevj:
Hmm. This is intended to help address this issue:
https://github.com/uProxy/uproxy/issues/1827

However, I'm struggling to observe much improvement. Since we've had a bunch of trouble with explicit channel IDs in the past, can you think of a good way to test this? My feeling is the slightly non-deterministic behaviour of the pool makes this hard to test - maybe it's worth tackling https://github.com/uProxy/uproxy/issues/1828 too?

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/252)

### DIFF
--- a/src/pool/pool.ts
+++ b/src/pool/pool.ts
@@ -79,18 +79,17 @@ class LocalPool {
     // TODO: limit the number of channels (probably should be <=256).
     if (this.pool_.length > 0) {
       var channel = this.pool_.shift();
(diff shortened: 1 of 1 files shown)


trevj:
As @bemasc pointed out to me in person this seems to have absolutely no effect, as you can see by opening `chrome://webrtc-internals`...

trevj:
OK, this doesn't work. Closing.

PR #253, opened 2015-08-19
Added new methods to arraybuffers

blanu:
Added split and convenience wrappers take and drop.
Added encodeShort and decodeShort.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/253)

### DIFF
--- a/src/arraybuffers/arraybuffers.ts
+++ b/src/arraybuffers/arraybuffers.ts
@@ -1,7 +1,7 @@
 // Byte-wise equality check of array buffers by comparison of each byte's
 // value.
 export function byteEquality(b1 :ArrayBuffer, b2 :ArrayBuffer)
-    : boolean {
+    :boolean {
   var a1 = new Uint8Array(b1);
   var a2 = new Uint8Array(b2);
   if(a1.byteLength !== a2.byteLength) return false;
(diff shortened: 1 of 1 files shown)


iislucas:
looking nice. Two small comments to think about.  :)

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[src/arraybuffers/arraybuffers.ts, line 123 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/253#-Jx65El0fJiwc2t7dH5c)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/bfb0ae300b901ac5da2014e9313ce9c7dedb7a4a/src/arraybuffers/arraybuffers.ts#L123)):</sup>
Probably better to do this as: 

```
export interface Pair<T> {
  first:T; 
  last: T; 
}
```
?

---

<sup>**[src/arraybuffers/arraybuffers.ts, line 137 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/253#-Jx65l0BGAyX066vUCvM)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/bfb0ae300b901ac5da2014e9313ce9c7dedb7a4a/src/arraybuffers/arraybuffers.ts#L137)):</sup>
Consider using slice here: more efficient? (I think it does a memcpy underneath)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/253)

PR #254, opened 2015-08-19
jpm-ify the firefox sample apps

trevj:
Turns out jpm does the signing thing, fixing our issues on Docker:
https://github.com/uProxy/uproxy/issues/1796

Seems to work just nicely now with uproxy-docker, here's the - finally! - complete throughput graph:

throughput

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/254)


bemasc:
:+1:

PR #255, opened 2015-08-20
Added AES implementation and tests

blanu:
This AES implementation was ported from Javascript to Typescript. Modes other than CBC were removed.
Tests were created using the output of the Javascript implementation to provide reference output.
The tasks to build and run the tests were added to the Gruntfile.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/255)


trevj:
I don't understand why you're copying code that's already available in NPM. For example, how will we keep up to date with bug fixes upstream?

I see you've added tests but they're probably better contributed upstream, right?

I doubt there's any reason why you can't use the NPM. We've been using the `crypto` NPM's implementation of `sha1` for ages without any issues. At first, we had to `browserify` the JS but once we switched to require it was even easier to import. Here's how we do it:
 1. [package.json entry](https://github.com/uProxy/uproxy-lib/blob/dev/package.json)
 1. [TypeScript definition](https://github.com/uProxy/uproxy-lib/blob/dev/third_party/sha1/sha1.d.ts)
 1. [import and call](https://github.com/uProxy/uproxy-lib/blob/dev/src/turn-frontend/messages.ts)

Happy to help you out here.

PR #256, opened 2015-08-21
make chrome app background scripts plain JS where possible

trevj:
Random Friday fixit...

The background script for those sample apps without any UI is dead simple: just a few lines of JS. If it's just JS we can avoid the need for a `browserify` step. adventure and uProbe already did this, and all our Firefox apps have always done this. Clearly, there is potential to template these files -- they differ only in the path to the freedomjs module.

Still warranted for sample apps with UI (copypaste, etc.).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/256)


trevj:
Update: do not use `=>` in regular JavaScript. Thanks!

bemasc:
:+1:

PR #258, opened 2015-08-24
timeout for probe candidate gathering

trevj:
Related issue:
https://github.com/uProxy/uproxy/issues/1795

My sense is we have one too many checks here which could be more cleanly implemented with a code restructure - chiefly, factoring out the probe connection into its own class - but I strongly suspect this will help us.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/258)

### DIFF
--- a/src/churn/churn.ts
+++ b/src/churn/churn.ts
@@ -144,9 +144,10 @@ export var filterCandidatesFromSdp = (sdp:string) : string => {
   //         https://github.com/uProxy/uproxy/issues/585
   export class Connection implements peerconnection.PeerConnection<ChurnSignallingMessage> {
 
-    // Maximum time to spend gathering ICE candidates:
-    //  https://github.com/uProxy/uproxy/issues/1795
-    private static PROBE_TIMEOUT_MS = 3000;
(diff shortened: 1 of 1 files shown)


src/churn/churn.ts
```diff
@@ -144,6 +144,10 @@ export var filterCandidatesFromSdp = (sdp:string) : string => {
   //         https://github.com/uProxy/uproxy/issues/585
   export class Connection implements peerconnection.PeerConnection<ChurnSignallingMessage> {
 
+    // Maximum time to spend gathering ICE candidates:
+    //  https://github.com/uProxy/uproxy/issues/1795
+    private static PROBE_TIMEOUT_MS = 3000;
```
bemasc: Trailing _ for private constants.  Also, if the real goal here is to avoid making the user wait too long, that might be worth documenting in a comment.

bemasc:
:+1:

src/churn/churn.ts
trevj: Good idea. I added the suffix and updated the comment.

PR #259, opened 2015-08-24
make AES algorithm available, via aes-js NPM

trevj:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/259)


blanu:
+1 looks good to me!

PR #260, opened 2015-08-24
Reduce heartbeat timeout to 15 seconds.

bemasc:
Now that uproxy has auto-reconnect, it's probably better to fail
sooner (e.g. to recover from an IP address change), and risk
occasional disconnections under extreme load.

This is intended to mitigate the Firefox disconnection issue: https://github.com/uProxy/uproxy/issues/1803

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/260)


trevj:
:+1:

PR #261, opened 2015-08-24
Added a transformer that encrypts packets using AES in CBC mode

blanu:
The encryption shaper uses AES in CBC mode.
The shared symmetric key is provided in configure().
Each packet includes a random Initialization Vector (IV).
This version does not yet implement packet fragmentation for packets where adding the initialization vector causes the length to exceed the expected MTU.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/261)

### DIFF
--- a/src/fancy-transformers/encryptionShaper.ts
+++ b/src/fancy-transformers/encryptionShaper.ts
@@ -51,6 +51,10 @@ export class EncryptionShaper implements Transformer {
   }
 
   public transform = (buffer:ArrayBuffer) :ArrayBuffer[] => {
+    // This transform performs the following steps:
+    // - Generate a new random 16-byte IV for every packet
+    // - Encrypt the packet contents with the random IV and symmetric key
+    // - Concatenate the IV and encrypted packet contents
     var iv :ArrayBuffer=this.makeIV_();
     var encrypted :ArrayBuffer=this.encrypt_(iv, buffer);
     var parts=[iv, encrypted]
@@ -58,6 +62,11 @@ export class EncryptionShaper implements Transformer {
   }
 
   public restore = (buffer:ArrayBuffer) :ArrayBuffer[] => {
+    // This restore performs the following steps:
+    // - Split the first 16 bytes from the rest of the packet
+    //     The two parts are the IV and the encrypted packet contents
+    // - Decrypt the encrypted packet contents with the IV and symmetric key
+    // - Return the decrypted packet contents
     var parts = arraybuffers.split(buffer, 16);
     var iv=parts[0];
     var ciphertext=parts[1];
--- a/src/fancy-transformers/encryptionShaper.ts
+++ b/src/fancy-transformers/encryptionShaper.ts
@@ -49,7 +49,7 @@ export class EncryptionShaper implements Transformer {
     // - Generate a new random 16-byte IV for every packet
     // - Encrypt the packet contents with the random IV and symmetric key
     // - Concatenate the IV and encrypted packet contents
-    var iv :ArrayBuffer=this.makeIV_();
+    var iv :ArrayBuffer=EncryptionShaper.makeIV();
     var encrypted :ArrayBuffer=this.encrypt_(iv, buffer);
     var parts=[iv, encrypted]
     return [arraybuffers.concat(parts)];
@@ -70,7 +70,7 @@ export class EncryptionShaper implements Transformer {
   // No-op (we have no state or any resources to dispose).
   public dispose = () :void => {}
 
-  private makeIV_ = () :ArrayBuffer => {
+  static makeIV = () :ArrayBuffer => {
     var randomBytes=new Uint8Array(16);
     crypto.getRandomValues(randomBytes);
     return randomBytes.buffer;
--- a/src/fancy-transformers/encryptionShaper.ts
+++ b/src/fancy-transformers/encryptionShaper.ts
@@ -1,0 +1,127 @@
+
+// TODO(bwiley): update uTransformers to be compatible with require
+// TODO(ldixon): update to a require-style inclusion.
+// e.g.
+//  import Transformer = require('uproxy-obfuscators/transformer');
+/// <reference path='../../../third_party/uTransformers/utransformers.d.ts' />
+/// <reference path='../../../third_party/typings/webcrypto/WebCrypto.d.ts' />
+/// <reference path='../../../third_party/aes-js/aes-js.d.ts' />
+
+import logging = require('../logging/logging');
+import arraybuffers = require('../arraybuffers/arraybuffers');
+import aes = require('aes-js');
+
+var log :logging.Log = new logging.Log('fancy-transformers');
+
+export interface EncryptionConfig {key:ArrayBuffer}
+export interface SerializedEncryptionConfig {key:string}
+
+// A packet shaper that encrypts the packets with AES CBC.
+export class EncryptionShaper implements Transformer {
+  private key_ :ArrayBuffer;
+
+  public constructor() {
+    log.info('Constructed encryption shaper');
+  }
+
+  // This method is required to implement the Transformer API.
+  // @param {ArrayBuffer} key Key to set, not used by this class.
+  public setKey = (key:ArrayBuffer) :void => {
+    // Do nothing.
+  }
+
+  public configure = (json:string) :void => {
+    log.debug("Configuring encryption shaper");
+
+    try {
+      var config=JSON.parse(json);
+
+      // Required parameter
+      if('key' in config) {
+        var encryptionConfig=this.deserializeConfig_(<SerializedEncryptionConfig>config);
+        this.key_=encryptionConfig.key;
+      } else {
+        log.error('Bad JSON config file');
+        log.error(json);
+        throw new Error("Encryption shaper requires key parameter");
+      }
+    } catch(err) {
+      // This is a common failure mode for transformers as any problem with the
+      // configuration usually results in an exception.
+      log.error("Transformer configure crashed");
+    }
+
+    log.debug("Configured encryption shaper");
+  }
+
+  public transform = (buffer:ArrayBuffer) :ArrayBuffer[] => {
+    var iv :ArrayBuffer=this.makeIV_();
+    var encrypted :ArrayBuffer=this.encrypt_(iv, buffer);
+    var parts=[iv, encrypted]
+    return [arraybuffers.concat(parts)];
+  }
+
+  public restore = (buffer:ArrayBuffer) :ArrayBuffer[] => {
+    var parts = arraybuffers.split(buffer, 16);
+    var iv=parts[0];
+    var ciphertext=parts[1];
+    return [this.decrypt_(iv, ciphertext)];
+  }
+
+  // No-op (we have no state or any resources to dispose).
+  public dispose = () :void => {}
+
+  private deserializeConfig_ = (config:SerializedEncryptionConfig) :EncryptionConfig => {
+    return {key:arraybuffers.hexStringToArrayBuffer(config.key)};
+  }
+
+  private makeIV_ = () :ArrayBuffer => {
+    var randomBytes=new Uint8Array(16);
+    crypto.getRandomValues(randomBytes);
+    return randomBytes.buffer;
+  }
+
+  private encrypt_ = (iv:ArrayBuffer, buffer:ArrayBuffer) :ArrayBuffer => {
+    var len :ArrayBuffer = arraybuffers.encodeShort(buffer.byteLength);
+    var remainder = (len.byteLength + buffer.byteLength) % 16;
+    var plaintext:ArrayBuffer;
+    if (remainder === 0) {
+      plaintext=arraybuffers.concat([len, buffer]);
+    } else {
+      var padding = new Uint8Array(16-remainder);
+      crypto.getRandomValues(padding);
+      plaintext=arraybuffers.concat([len, buffer, padding.buffer]);
+    }
+
+    var cbc = new aes.ModeOfOperation.cbc(new Uint8Array(this.key_), new Uint8Array(iv));
+    var chunks = arraybuffers.chunk(plaintext, 16);
+    for(var x = 0; x < chunks.length; x++) {
+      var plainChunk=arraybuffers.arrayBufferToBuffer(chunks[x]);
+      var cipherChunk=cbc.encrypt(plainChunk);
+      chunks[x]=arraybuffers.bufferToArrayBuffer(cipherChunk);
+    }
+
+    return arraybuffers.concat(chunks);
+  }
+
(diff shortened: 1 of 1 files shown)


trevj:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 8 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_B7MOtf9NTGX3l_Lo)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L8)):</sup>
You only need the `aes-js` and `webcrypto` references, right?

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 12 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_BF1SW8eOLI-sziA3)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L12)):</sup>
Not in the style guide (yet) but probably good to sort imports alphabetically.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 14 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_GqIoT5ZW4digpWww)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L14)):</sup>
You'll be adding other transformers to this directory, right? I think the class name, e.g. `encryption-shaper`, would be better.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 24 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_BkuKpi0VG12MrnY_)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L24)):</sup>
If the constructor is only going to log this, I would remove it altogether (unless you've found it useful for debugging...hopefully not).

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 29 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_Bs3YsHYRoFs7-Auu)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L29)):</sup>
I would dispense with the jsdoc and make this a one-liner like `dispose` below, i.e.:
```typescript
// No-op (we have no key).
public setKey(key:ArrayBuffer) : void => {}
```

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 40 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_CpzR6cNj01kljDrO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L40)):</sup>
It looks like the supplied object has one field, named `key`. Do you imagine other fields being added over time? If not, did you consider implementing only `setKey` instead? :-)

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 57 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_EN4mgSLcRLUpSZuQ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L57)):</sup>
I think it would be helpful to explain that a new IV is generated for every packet and that the transformed packet is the concatenation of an IV, the length of the encoded bytes, and the AES-ified bytes.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261)

blanu:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 8 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_B7MOtf9NTGX3l_Lo)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L8)):</sup>
A reference to utransformers is also required in order to get the Transformer interface.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 12 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_BF1SW8eOLI-sziA3)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L12)):</sup>
I will make this change AND add it to the style guide.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 24 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_BkuKpi0VG12MrnY_)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L24)):</sup>
I removed the logging. The constructor function is still necessary though for typechecking to succeed. All classes implementing the Transformer interface must now have a constructor function, even if it's empty.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 40 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_CpzR6cNj01kljDrO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L40)):</sup>
I did consider this. There are valid arguments for both approaches. The reason I chose to pass the key into configure() instead of setKey() is that it is more consistent across the set of transformers. There are essentially two configuration styles present here. One is to have a setX() method for each parameter, the other is to pass a JSON-encoded map of parameters into configure(). The disadvantage of the setX approach is that the Transformer interface cannot be extended to support all of the parameters for every transformer. For this particular transformer, there is indeed a key parameter, but the other transformers do not have keys. The disadvantage of the JSON approach is that JSON strings cannot include ArrayBuffers and several of the transformers require ArrayBuffers for configuration. So these must be serialized and deserialized as strings.

I would like to propose a third option, which is to eliminate setKey() in favor of configure(), but to change the type signature of configure from configure(string):void to configure(Object):void. Each transformer can export an interface for its configuration object, as I did with EncryptionConfig here. This can be used for typechecking on both the caller and inside the transformer. This is essentially how I am doing things now, but changing the type signature means that we avoid serializing and deserializing into JSON and also the extra step of serializing and deserializing the ArrayBuffers into strings.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261)

trevj:
Review status: 0 of 1 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 8 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_B7MOtf9NTGX3l_Lo)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L8)):</sup>
Ah, of course.

But you still don't need the first four lines?

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 12 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_BF1SW8eOLI-sziA3)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L12)):</sup>
Awesome, thanks.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 16 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jxafh94y6zpGeWp4scY)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9d28b2d5d1f6363b8b19faefd2508182e7240b71/src/fancy-transformers/encryptionShaper.ts#L16)):</sup>
Why have both? Seems like once you've deserialized the config you can just set the member variable `key_`.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 36 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jx_CpzR6cNj01kljDrO)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/dc11efd72f464eb875c0001787f6539115d16319/src/fancy-transformers/encryptionShaper.ts#L36)):</sup>
Aha. I just realised `setKey` accepts an `ArrayBuffer` so my suggestion isn't as neat as I thought. Sorry.

OK, let's leave it as a JSON object. In that case, I have a few other suggestions to simplify this method...

(BTW, IIRC, we chose JSON over `Object` precisely because it's serializable, and can traverse freedomjs module boundaries, etc. -- let's leave as-is for now)

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 36 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-JxaewzzP3P3nZ4IuX-p)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9d28b2d5d1f6363b8b19faefd2508182e7240b71/src/fancy-transformers/encryptionShaper.ts#L36)):</sup>
I would have `deserializeConfig_` throw any errors, e.g. if the JSON is unparseable or is missing a required key. Then, this method will simply propagate the errors (and is `deserializeConfig_` called anywhere else? if not, you should probably just defactor it into this function).

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 44 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-JxafsfQyeD5NnoTGhSM)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9d28b2d5d1f6363b8b19faefd2508182e7240b71/src/fancy-transformers/encryptionShaper.ts#L44)):</sup>
FYI, we usually use `e` for catch blocks.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 58 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jxag5YH3dyEwTNI0c_W)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9d28b2d5d1f6363b8b19faefd2508182e7240b71/src/fancy-transformers/encryptionShaper.ts#L58)):</sup>
style: space before and after assignment operator (and throughout)

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 79 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-JxafM1AlM0XgOPcMpfC)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9d28b2d5d1f6363b8b19faefd2508182e7240b71/src/fancy-transformers/encryptionShaper.ts#L79)):</sup>
This function can be static.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 83 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-JxafOkPqbE6scZ4VSzk)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/9d28b2d5d1f6363b8b19faefd2508182e7240b71/src/fancy-transformers/encryptionShaper.ts#L83)):</sup>
This function can be static.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261)

trevj:
So close!

Some minor style things and `configure` needs some love, then :+1:.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 23 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jxft4gOJH8mpDj8Xw-N)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/921b3307631b358845a17f41b880984713872076/src/fancy-transformers/encryptionShaper.ts#L23)):</sup>
So this method can be a lot simpler, and you're swallowing the exception in case of badly-formatted JSON.

Suggest:
```typescript
// |json| should be a serialized EncryptionConfig instance.
// Throws an error if the supplied JSON is invalid or is missing a required field.
public configure = (json:string) : void => {
  var config = JSON.parse(json);
  if (!('key' in config)) {
    throw new Error('Encryption shaper requires key parameter');
  }
  var encryptionConfig = <EncryptionConfig>config;
  this.key_ = arraybuffers.hexStringToArrayBuffer(encryptionConfig.key);
}
```

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 30 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-JxfuY-9GPAkPQUs5gtl)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/921b3307631b358845a17f41b880984713872076/src/fancy-transformers/encryptionShaper.ts#L30)):</sup>
style: space after if (and a few other places throughout)

```typescript
if ('key' in config) {
...
}
```

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261)

blanu:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 29 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-Jxft4gOJH8mpDj8Xw-N)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/921b3307631b358845a17f41b880984713872076/src/fancy-transformers/encryptionShaper.ts#L29)):</sup>
Okay I have made this change.

---

<sup>**[src/fancy-transformers/encryptionShaper.ts, line 36 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261#-JxfuY-9GPAkPQUs5gtl)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/921b3307631b358845a17f41b880984713872076/src/fancy-transformers/encryptionShaper.ts#L36)):</sup>
Okay I have made this change.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/261)

PR #263, opened 2015-08-25
specify a channel ID, to effectively double the number of possible channels

trevj:
pre-:+1:-ed earlier today by @bemasc, who figured out the fix to `churn.ts`.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/263)


trevj:
Related issue:
https://github.com/uProxy/uproxy/issues/1827

PR #265, opened 2015-08-27
add methods to freedom social api

salomegeo:
Does this hurt anything?

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/265)

### DIFF
--- a/third_party/freedom-typings/social.d.ts
+++ b/third_party/freedom-typings/social.d.ts
@@ -35,7 +35,8 @@
       url          ?:string;
       // Image URI (e.g. data:image/png;base64,adkwe329...)
       imageData    ?:string;
-      timestamp    ?:number;
+      lastUpdated  ?:number;
+      status       ?:number;
     }
 
     interface Users   { [userId:string]   : UserProfile; }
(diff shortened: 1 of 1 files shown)


trevj:
I'm not familiar with this interface but now that the typings have moved to DefinitelyTyped (https://github.com/uProxy/uproxy-lib/pull/273) you will probably need to make the changes there, first.

@jpevarnek should be able to help you out.

jpevarnek:
The DefinitelyTyped contribution guidelines are at http://definitelytyped.org/guides/contributing.html

It looks like this makes our typings incompatible with the Freedom interface (https://github.com/UWNetworksLab/freedom/blob/master/interface/social.js), is that intentional?

jpevarnek:
Hey @salomegeo, just wanting the check if there was an update on this.  Thanks!

trevj:
*click*

This adds the new methods in the social2 API. This sure would be handy for the cloud provider I've been working on recently but, yes, I think the appropriate place for this now is DefinitelyTyped.

Considering that and how I'm hoping to make `master` the default branch, I'm going to go ahead and close this unless you've got major objections (just re-open on `master`, when it's ready).

PR #266, opened 2015-08-27
Added a class to model packet fragments

blanu:
The fragment class models a packet fragment for fragmentation and defragmentation
It has methods to serialize and deserialize the packet fragment format.
It also has a method for generating a random, unique fragment identifier.
Three utility methods are include: takeByte_, dropByte_, and encodeByte_, in order to make the serialization and deserialization code more readable.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/266)

### DIFF
--- a/src/arraybuffers/arraybuffers.ts
+++ b/src/arraybuffers/arraybuffers.ts
@@ -150,6 +150,18 @@ export function split(buffer:ArrayBuffer, firstLen:number) :[ArrayBuffer, ArrayB
   return [first, last];
 }
 
+// Wrap a byte in an ArrayBuffer
+export function encodeByte(num:number) :ArrayBuffer {
+  var bytes = new Uint8Array(1);
+  bytes[0] = num;
+  return bytes.buffer;
+}
+
+// Convert an ArrayBuffer with one element to a byte
+export function decodeByte(buffer:ArrayBuffer) :number {
+  return new Uint8Array(buffer)[0];
+}
+
 // Takes a number and returns a two byte (network byte order) representation
 // of this number.
 export function encodeShort(len:number) :ArrayBuffer {
@@ -165,4 +177,28 @@ export function encodeShort(len:number) :ArrayBuffer {
   var bytes = new Uint8Array(buffer);
   var result = (bytes[0] << 8) | bytes[1];
   return result;
+}
+
+// Parse a buffer into parts using an array of lengths for each part.
+// Any remaining bytes left over after parsing are appending to the end.
+// This function is intended for using in parsing binary protocols with
+// fixed length fields.
+export function parse(buffer:ArrayBuffer, lengths:number[]) :ArrayBuffer[] {
+  var parts :ArrayBuffer[] = [];
+
+  var part :ArrayBuffer = null;
+  var remaining :ArrayBuffer = buffer;
+
+  lengths.forEach(length => {
+    if (remaining.byteLength >= length) {
+      [part, remaining] = split(remaining, length);
+      parts.push(part);
+    } else {
+      throw new Error("Not enough bytes to parse");
+    }
+  });
+
+  parts.push(remaining);
+
+  return parts;
 }
--- a/src/fancy-transformers/fragment.ts
+++ b/src/fancy-transformers/fragment.ts
@@ -1,12 +1,10 @@
-/// <reference path='../../../third_party/typings/webcrypto/WebCrypto.d.ts' />
-
 import arraybuffers = require('../arraybuffers/arraybuffers');
 import logging = require('../logging/logging');
 
 var log :logging.Log = new logging.Log('fancy-transformers');
 
 // A Fragment represents a piece of a packet when fragmentation has occurred.
-interface Fragment {
+export interface Fragment {
   length :number;
   id :ArrayBuffer;
   index :number;
@@ -35,7 +33,7 @@ export function makeRandomId() :ArrayBuffer {
 //   - padding, variable number of bytes, whatever is left after the payload
 export function decode(buffer:ArrayBuffer) :Fragment {
   var [lengthBytes, fragmentId, fragmentNumber, totalNumber, remaining] =
-    arraybuffers.parse([2, 32, 1, 1], buffer);
+    arraybuffers.parse(buffer, [2, 32, 1, 1]);
   var length = arraybuffers.decodeShort(lengthBytes);
 
   var payload :ArrayBuffer = null;
@@ -51,13 +49,13 @@ export function decode(buffer:ArrayBuffer) :Fragment {
     throw new Error("Fragment could not be decoded, shorter than length");
   }
 
-  return new Fragment(
-    length,
-    fragmentId,
-    decodeByte(fragmentNumber),
-    decodeByte(totalNumber),
-    payload,
-    padding);
+  return {
+    length: length,
+    id: fragmentId,
+    index: arraybuffers.decodeByte(fragmentNumber),
+    count: arraybuffers.decodeByte(totalNumber),
+    payload: payload,
+    padding: padding};
 }
 
 // Serialize a Fragment object so that it can be sent as a packet
@@ -72,8 +70,8 @@ export function encode(fragment:Fragment) :ArrayBuffer {
   return arraybuffers.concat([
     arraybuffers.encodeShort(fragment.length),
     fragment.id,
-    encodeByte_(fragment.index),
-    encodeByte_(fragment.count),
+    arraybuffers.encodeByte(fragment.index),
+    arraybuffers.encodeByte(fragment.count),
     fragment.payload,
     fragment.padding
   ]);
--- a/src/fancy-transformers/fragment.ts
+++ b/src/fancy-transformers/fragment.ts
@@ -5,123 +5,80 @@ import logging = require('../logging/logging');
 
 var log :logging.Log = new logging.Log('fancy-transformers');
 
-// Return the first byte of an ArrayBuffer
-var takeByte_ = (buffer:ArrayBuffer) :number => {
+// A Fragment represents a piece of a packet when fragmentation has occurred.
+interface Fragment {
+  length :number;
+  id :ArrayBuffer;
+  index :number;
+  count :number;
+  payload :ArrayBuffer;
+  padding :ArrayBuffer;
+}
+
+// Make a random 32-byte identifier for the packet fragment.
+// This should be unique, or else defragmentation breaks.
+// A 32-byte size was chosen as this is a common hash function output size.
+// In the future, a hash could perhaps be used instead of a random identifier.
+export function makeRandomId() :ArrayBuffer {
+  var randomBytes = new Uint8Array(32);
+  crypto.getRandomValues(randomBytes);
+  return randomBytes.buffer;
+}
+
+// Deserialize the content of a packet into a Fragment object
+// The Fragment format is as follows:
+//   - length of the payload, 2 bytes
+//   - id, 32 bytes
+//   - fragment number, 1 byte
+//   - total number of fragments for this id, 1 byte
+//   - payload, number of bytes specified by length field
+//   - padding, variable number of bytes, whatever is left after the payload
+export function decode(buffer:ArrayBuffer) :Fragment {
+  var [lengthBytes, fragmentId, fragmentNumber, totalNumber, remaining] =
+    arraybuffers.parse([2, 32, 1, 1], buffer);
+  var length = arraybuffers.decodeShort(lengthBytes);
+
+  var payload :ArrayBuffer = null;
+  var padding :ArrayBuffer = null;
+
+  if (remaining.byteLength > length) {
+    [payload, padding] = arraybuffers.split(remaining, length);
+  } else if (buffer.byteLength === length) {
+    payload = remaining;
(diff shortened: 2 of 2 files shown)


trevj:
Review status: 0 of 1 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

<sup>**[src/fancy-transformers/fragment.ts, line 26 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxksfBz2girM6O7pUpT)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L26)):</sup>
Public fields are rarely a good idea, although I guess in certain cases - such as a `struct`-style object like this - they have a purpose.

Is mutability important in your code, e.g. for performance reasons? Generally immutability makes it much much easier to reason about your code and public fields make immutability impossible.

---

<sup>**[src/fancy-transformers/fragment.ts, line 35 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxklcEhGnZiVkfZAMPI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L35)):</sup>
TypeScript has a neat facility for declaring member variables in constructors, so you could re-write the opening of this class as:
```typescript
// A Fragment represents a piece of a packet when fragmentation has occurred.
class Fragment {
  public constructor(
    private length_:number,
    private id_:ArrayBuffer,
    private index_:number,
    private count_:number,
    private payload_:ArrayBuffer,
    private padding_:ArrayBuffer) {}

// ...
// ...
}
```
(works for public fields, too)

---

<sup>**[src/fancy-transformers/fragment.ts, line 53 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxksP9tH-fRdOBXB5GU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L53)):</sup>
This isn't called anywhere here, just wanted to doublecheck you'll want it in future?

---

<sup>**[src/fancy-transformers/fragment.ts, line 61 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxkudJr-eIw1x3qu1XR)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L61)):</sup>
I like the description of the `Fragment` bytes, I would add here that length refers to the length of `payload`.

---

<sup>**[src/fancy-transformers/fragment.ts, line 67 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-Jxkv7RB5nX1UdcBped3)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L67)):</sup>
I think you should make this a top-level, exported function and drop `fragment` from the title.

That way, it'll read much better when called, e.g.
```typescript
import fragment = require('../fragment/fragment');
var buffer = ; // a load of bytes from the network
var fragment = fragment.decode(buffer);
```

---

<sup>**[src/fancy-transformers/fragment.ts, line 76 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxktUr6GBUpikbRQhv0)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L76)):</sup>
I see what you're doing but the meaning of buffer changes several times during the course of this method, imposing an unnecessary load on the reader.

I think it would be easier to read if you just accessed the bytes of some `Uint8Array`, e.g. which would you prefer to come across:
```typescript
var fragmentNumber = someMethodINeverHeardOf_(buffer);
buffer = someOtherMethodINeedToKnowAbout_(buffer);
var totalNumber = takeByte_(buffer);
```
vs.
```typescript
var fragmentNumber = bytes[0];
var totalNumber = bytes[1];
```
?

---

<sup>**[src/fancy-transformers/fragment.ts, line 104 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxkuukF2tTbF6dd6Q3_)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L104)):</sup>
You can condense this into:
```typescript
return new Fragment(
    length,
    fragmentId,
    fragmentNumber,
    totalNumber,
    payload,
    padding);
```

---

<sup>**[src/fancy-transformers/fragment.ts, line 115 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxkvhUlx2FNYAFqIBEQ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L115)):</sup>
As for decode, I would make this top-level and rename it `encode`.

---

<sup>**[src/fancy-transformers/fragment.ts, line 127 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxkmYR1jvDgHkWwLs47)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L127)):</sup>
You can just prefix the class definition above with `export`.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266)

trevj:
Review status: 0 of 1 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

<sup>**[src/fancy-transformers/fragment.ts, line 67 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-Jxkv7RB5nX1UdcBped3)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L67)):</sup>
Aaaaand of course that doesn't work...how about:

```typescript
import fragments = require('../fragment/fragment');
var buffer = ; // a load of bytes from the network
var fragment = fragments.decode(buffer);
```

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266)

blanu:
Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

<sup>**[src/fancy-transformers/fragment.ts, line 26 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxksfBz2girM6O7pUpT)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L26)):</sup>
As discussed in person, I will rewrite this class as a interface.

---

<sup>**[src/fancy-transformers/fragment.ts, line 35 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxklcEhGnZiVkfZAMPI)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L35)):</sup>
No longer needed as I am switching to a struct, but cool to know for the future.

---

<sup>**[src/fancy-transformers/fragment.ts, line 53 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxksP9tH-fRdOBXB5GU)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L53)):</sup>
This is called elsewhere when making a new Fragment object. I just thought I should keep it in this module as that allows for fragment module to determine the semantics of the identifier.

---

<sup>**[src/fancy-transformers/fragment.ts, line 61 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxkudJr-eIw1x3qu1XR)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L61)):</sup>
Okay I will make this change.

---

<sup>**[src/fancy-transformers/fragment.ts, line 67 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-Jxkv7RB5nX1UdcBped3)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L67)):</sup>
Okay I will make this change. As I moved from a class to an interface, I similarly make the encodeFragment() method into an exported function called encode() that takes a Fragment as an argument.

---

<sup>**[src/fancy-transformers/fragment.ts, line 76 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxktUr6GBUpikbRQhv0)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L76)):</sup>
Okay, I have rewritten. That was the Haskell-ish way of parsing a buffer, where take and drop are well-known idioms. It might not be the most appropriate for Typescript.

I have rewritten it to use the new Typescript 1.5 array destructuring and a new method I added to arraybuffers that will split up a buffer with fixed width fields into a list suitable for destructuring.

---

<sup>**[src/fancy-transformers/fragment.ts, line 104 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxkuukF2tTbF6dd6Q3_)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L104)):</sup>
Okay I will make this change.

---

<sup>**[src/fancy-transformers/fragment.ts, line 115 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxkvhUlx2FNYAFqIBEQ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L115)):</sup>
Okay I made this change.

---

<sup>**[src/fancy-transformers/fragment.ts, line 127 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxkmYR1jvDgHkWwLs47)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L127)):</sup>
I made a similar change. I exported the interface and the functions.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266)

trevj:
Cool. Just one question, then I think we're good to go. :+1:

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/fancy-transformers/fragment.ts, line 0 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxksfBz2girM6O7pUpT)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/eb45506ea1599c37b4c345c2078cf9409d2f66b5/src/fancy-transformers/fragment.ts#L0)):</sup>
Great idea, I like this a lot better.

---

<sup>**[src/fancy-transformers/fragment.ts, line 13 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxpzwNi1xoq7aumQG40)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/7b5ee157e336e183f26f32396efb1ef874f56dd4/src/fancy-transformers/fragment.ts#L13)):</sup>
I do wonder if it's really necessary to model `padding`, too?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266)

trevj:
Still :+1: !

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/fancy-transformers/fragment.ts, line 13 \[r3\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266#-JxpzwNi1xoq7aumQG40)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/7b5ee157e336e183f26f32396efb1ef874f56dd4/src/fancy-transformers/fragment.ts#L13)):</sup>
OK. I suspect we could make the padding a little more transparent (unless clients care about the contents of the padding, which I guess is possible) but I don't have a good suggestion right now!

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/266)

PR #166, opened 2015-05-07
merge in uproxy-networking code

trevj:
This copies across all code from the [uproxy-networking repository](https://github.com/uProxy/uproxy-networking) into this one, with the intention of deleting that repository and associated NPM. Fixes:
https://github.com/uProxy/uproxy/issues/1421

I tried to copy as much as possible, with the following exceptions:
- update paths, in code and JSON files
- workaround TypeScript compiler errors owing to latest es6-promise definitions
- fix some sample apps that had atrophied, viz. Simple TURN, and churn sample apps

This is a huge diff so you might prefer to patch this in and perform a diff locally; I prefer `meld` and if your system is anything like mind you can execute a command similar to this one to highlight the changes in this pull request:

``` bash
meld ~/src/uproxy-networking ~/src/uproxy-lib
```

The code is copied at commit (ce0efe8)[https://github.com/uProxy/uproxy-networking/commit/ce0efe8ae8e9e506b0e13a41c30438866ce59c78], which means the following pull requests are outstanding:
- ~~https://github.com/uProxy/uproxy-networking/pull/257~~
- https://github.com/uProxy/uproxy-networking/pull/255
- ~~https://github.com/uProxy/uproxy-networking/pull/245~~
- ~~https://github.com/uProxy/uproxy-networking/pull/226~~
- https://github.com/uProxy/uproxy-networking/pull/210
- ~~https://github.com/uProxy/uproxy-networking/pull/259~~
- ~~https://github.com/uProxy/uproxy-networking/pull/258~~

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/166)

PR #267, opened 2015-08-27
migrate to TypeScript 1.5

trevj:
Related issue:
https://github.com/uProxy/uproxy/issues/1477

This is for selfish reasons -- I desperately needed ArcticTypeScript back and it now only compiles projects with TS1.5+.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/267)

### DIFF
--- a/src/socks-common/socks-headers.ts
+++ b/src/socks-common/socks-headers.ts
@@ -176,7 +176,11 @@ export function composeAuthHandshakeBuffer(auths:Auth[]) : ArrayBuffer {
   var handshakeBytes = new Uint8Array(auths.length + 2);
   handshakeBytes[0] = VERSION5;
   handshakeBytes[1] = auths.length;
-  handshakeBytes.set(auths, 2);
+  // TODO: Revert to set() when Uint8Array's type definition is fixed in TS1.6:
+  //         https://github.com/Microsoft/TypeScript/issues/3979
+  for (var i = 0; i < auths.length; i++) {
+    handshakeBytes[i + 2] = auths[i];
+  }
   return handshakeBytes.buffer;
 }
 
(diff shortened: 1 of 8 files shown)


trevj:
I went for bonus points and removed the workarounds introduced in https://github.com/uProxy/uproxy-lib/pull/166.

src/socks-common/socks-headers.ts
```diff
@@ -176,7 +176,11 @@ export function composeAuthHandshakeBuffer(auths:Auth[]) : ArrayBuffer {
   var handshakeBytes = new Uint8Array(auths.length + 2);
   handshakeBytes[0] = VERSION5;
   handshakeBytes[1] = auths.length;
-  handshakeBytes.set(auths, 2);
+  // TODO: Revert to set() when Uint8Array's type definition is fixed in TS1.6:
```
iislucas: I think it would be better to use a type-cast here, as you don't have to have different compiled JS code, e.g. : 
```
(<any>handshakeBytes).set(auths, 2);
```
Or you might explore if you can create a local over-ride at the top of the file. 
```
interface Uint8Array {
  set(auths:Auth[], n:number);
}
```
Same for other uses of set.

iislucas:
nice! I think we can do better than rewriting set, but apart from that it looks good.

src/socks-common/socks-headers.ts
trevj: Right. I couldn't find a way to augment the `Uint8Array` definition so I took your first suggestion.

trevj:
Thanks! Updated...PTAL.

iislucas:
:+1:

PR #268, opened 2015-08-28
Move freedom-related typings to a single file

jpevarnek:
This moves the freedom-related typings to a single file, freedom.d.ts,
for easier inclusion and in preparation for merging them into Definitely
Typed.

For freedom modules, this using separate typescript modules (e.g.
freedom.Console) in order to encapsulate everything that is needed for
that module.  The actual module (i.e. what is returned by
freedom['core.console']) is then located under freedom.Console.Console.

This also adds direct typing of the different calls to known module
providers (e.g. freedom['core.console']() now is typed as
freedom.Console.Console instead of any)

The types themselves should be relatively unchanged, however, they now
live in a cleaner space.

The core and module environments still have separate files to expose the
global variable (a Union type did not work for that...I tried).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/268)


trevj:
Super. Thanks for doing this. I patched it in and everything seems to run fine. It will be great to have these typings farmed out to DefinitelyTyped.

Just two comments:
 1. We generally indent by two spaces, rather than the four you've chosen.
 1. Port control isn't part of the "standard" freedomjs set of modules. I don't really care but just FYI in case you missed it (I guess social isn't, either, so you've just chosen a set of common freedomjs modules?)

jpevarnek:
1. Yeah, I was debating on that one a bit.  The DefinitelyTyped guide (http://definitelytyped.org/guides/contributing.html) says that you should use either four spaces or tabs for indentation.  I think I probably could have gotten it in with two-space indentation, but I wanted to follow their conventions as much as possible.

2. I went with whatever module typings we had in uproxy-lib and used.  If Freedom gets big enough that there begin to be name collisions between modules, that might become a problem, but I don't think it is a problem yet.

trevj:
1. Got it. That didn't even occur to me.
 1. Agreed. I don't feel strongly about it and don't see a problem.

:+1:

PR #269, opened 2015-08-28
Use ID 0 for the control channel, to avoid collision with the pool.

bemasc:
This fixes https://github.com/uProxy/uproxy/issues/1859

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/269)


trevj:
:+1: 

Thanks!

PR #271, opened 2015-08-28
batching and compression for signalling messages

trevj:
Related issue:
https://github.com/uProxy/uproxy/issues/1490

This introduces `SignalBatcher`, which can be used to batch and compress signalling channel messages. The uProxy frontend should be able to use this, removing a bunch of tricky logic from the frontend.

I'm seeing a great reduction in size:

description | size (bytes) | percent reduction
--- | --- | ---
unbatched signalling channel messages (JSON) | 1208 | 0%
batched messages (JSON) | 1104 | 8.6%
gzipped messages (binary) | 589 | 51.2%

Since there's a lot of redundancy in signalling messages (e.g. the word candidate like 50 times) I would not expect the final result to grow much beyond ~1000 characters even when many more STUN servers are queried. That's well under the ~2000 that starts to pose a problem in Hangouts.

Summary of changes:
 * have churn emit a `NO_MORE_CANDIDATES` signal when it's done (should be harmless to existing clients)
 * `SignalBatcher` can be attached to your signalling channel queue and it groups by `NO_MORE_CANDIDATES` messages
 * update copypaste sample to use `SignalBatcher`

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/271)


bemasc:
Review status: 0 of 13 files reviewed at latest revision, 8 unresolved discussions, some commit checks pending.

---

<sup>**[src/bridge/onetime.ts, line 36 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3PSeY9jkS-FOT1U8f)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/bridge/onetime.ts#L36)):</sup>
Would this to be better off as part of the Bridge interface?

---

<sup>**[src/bridge/onetime.ts, line 87 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3ORVImlc0bLmt4wkw)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/bridge/onetime.ts#L87)):</sup>
This seems a little scary to me, because it loses all other `providerName`s after the first.  Maybe it would be better to make this use `for (var providerName in message.signals)`?

---

<sup>**[src/bridge/onetime.ts, line 94 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3OfMWnaOrpwTWSVum)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/bridge/onetime.ts#L94)):</sup>
This also seems scary, if there are already multiple signals in one message.  I would prefer a loop.

---

<sup>**[src/bridge/onetime.ts, line 100 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3P5jO-GmGkKNzl3Sv)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/bridge/onetime.ts#L100)):</sup>
Doing this as a callback is pretty unusual for us.  Why not a Promise ... or just stay synchronous and use the return value of `addToBatch`?

---

<sup>**[src/bridge/onetime.ts, line 130 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3PoIshepSL2fpSlgh)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/bridge/onetime.ts#L130)):</sup>
Deserves a comment.  Is this just to try to hint the garbage collector?  (That shouldn't be necessary...)

---

<sup>**[src/copypaste-socks/freedom-module.ts, line 105 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3SJIiAdQ_TN_3CIUR)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/copypaste-socks/freedom-module.ts#L105)):</sup>
`encodedMessage` is a `string`, though, right?  So why call `toString`?

---

<sup>**[src/copypaste-socks/polymer-components/getter.html, line 33 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3RkE6WXwojwwVVyEG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/copypaste-socks/polymer-components/getter.html#L33)):</sup>
Numeric comparison on a string?

---

<sup>**[src/copypaste-socks/polymer-components/getter.ts, line 21 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3SRGmuMRPNQjriW47)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/copypaste-socks/polymer-components/getter.ts#L21)):</sup>
Seems like maybe an error should be thrown here?  Silently ignoring the input seems bad.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271)

trevj:
Review status: 0 of 11 files reviewed at latest revision, 8 unresolved discussions, some commit checks pending.

---

<sup>**[src/bridge/onetime.ts, line 36 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3PSeY9jkS-FOT1U8f)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/bridge/onetime.ts#L36)):</sup>
Right now, nothing in `Bridge` needs this. I'm reluctant to push it "down" (to a more important class) until there's a clear need. Good question, though.

---

<sup>**[src/bridge/onetime.ts, line 87 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3ORVImlc0bLmt4wkw)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/bridge/onetime.ts#L87)):</sup>
It's safe given that `addToBatch` raises an exception if called with signals having >1 `providerName`. That said, it's straightforward to add some checks. Performance isn't exactly key here, so I extracted out the code to check that messages just have provider and call it here, too.

---

<sup>**[src/bridge/onetime.ts, line 94 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3OfMWnaOrpwTWSVum)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/bridge/onetime.ts#L94)):</sup>
See above.

---

<sup>**[src/bridge/onetime.ts, line 100 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3P5jO-GmGkKNzl3Sv)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/bridge/onetime.ts#L100)):</sup>
I'm keeping in mind things like renegotiation, which could lead to a new batch of signalling messages (I'm sure we don't support renegotiation well today but if we ever did move to use this `SignalBatcher` for every-day peerconnection negotiation, it would become important).

A `Promise` approach won't work if there can be multiple batches, and using `addToBatch`'s return value is problematic because of `PeerConnection`'s use of handler queues.

---

<sup>**[src/bridge/onetime.ts, line 130 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3PoIshepSL2fpSlgh)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/bridge/onetime.ts#L130)):</sup>
Not GC, this is in anticipation of another batch (see comment above).

I did move the statement and a comment in the interest of clarity, though.

---

<sup>**[src/copypaste-socks/freedom-module.ts, line 105 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3SJIiAdQ_TN_3CIUR)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/copypaste-socks/freedom-module.ts#L105)):</sup>
Right, good catch. Fixed.

---

<sup>**[src/copypaste-socks/polymer-components/getter.ts, line 21 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3SRGmuMRPNQjriW47)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/copypaste-socks/polymer-components/getter.ts#L21)):</sup>
Right...I've re-introduced the `consumeInboundText` function which sends the box content to the freedom module for validation before enabling the button. Should be a lot better.

---

<sup>**[src/copypaste-socks/polymer-components/getter.html, line 33 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3RkE6WXwojwwVVyEG)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/copypaste-socks/polymer-components/getter.html#L33)):</sup>
Oops! Not relevant now that I've introduced a round-trip to the freedom module for validating the input box's contents (see below).

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271)

bemasc:
Review status: 0 of 11 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/copypaste-socks/polymer-components/getter.ts, line 23 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3SRGmuMRPNQjriW47)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/copypaste-socks/polymer-components/getter.ts#L23)):</sup>
Hmm.  `parseInboundMessages` takes a mandatory `string` argument.  If it's actually not mandatory, maybe its type signature should be changed to make that an optional argument?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271)

trevj:
Review status: 0 of 11 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/copypaste-socks/polymer-components/getter.ts, line 23 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271#-Jy3SRGmuMRPNQjriW47)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/994155f0b6e948cefb62a521e433455caa222b5e/src/copypaste-socks/polymer-components/getter.ts#L23)):</sup>
I did update the function's signature in `copypaste-api.ts`, can you take another look?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/271)

bemasc:
:+1:

PR #272, opened 2015-08-31
TURN server cleanup and refactor

trevj:
A few things...call this a preview of how I'd like the SOCKS server to turn out:
 * Move all TURN code to a new top-level package, `turn` (this code predates our use of `require`).
 * Remove the frontend and backend freedomjs modules (we used to give essentially every class its own module, which is way excessive).
 * Rather than an `ArrayBuffer` for frontend to backend communication, use a pair of `StunMessage` and `Endpoint`.

I'm actually most excited by the last point. It's going to allow us consolidate serialisation code - e.g. for a distributed WebRTC implementation - in one place, rather than duplicating it in both frontend and backend as we do today. For now, this makes for a way more efficient Simple TURN sample app.

I strongly considered having the `Frontend` and `Backend` inherit from a superclass but now the amount of shared code would so low I don't think it's worth it.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/272)


bemasc:
:+1: and thanks for taking on the next wave of architecture planning discussions.

PR #275, opened 2015-09-01
add a limit on the number of channels, with plumbing back to the SOCKS client

trevj:
Related issue:
https://github.com/uProxy/uproxy/issues/1815

As discussed, this makes `PeerConnection` search the SDP for the maximum number of supported channels and, if found, cap the number of open channels to that maximum. Additionally, this (finally!) sends error bytes to the client when a channel cannot be obtained.

Informally, I can say this _seems_ to finally solve the "infinite loading" problem on complex sites such as time.com.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/275)

### DIFF
--- a/src/webrtc/peerconnection.ts
+++ b/src/webrtc/peerconnection.ts
@@ -488,6 +488,11 @@ export class PeerConnectionClass implements PeerConnection<signals.Message> {
           this.maxChannels_ + ')'));
     }
 
+    if (options && options.id && options.id >= this.maxChannels_) {
+      return Promise.reject(new Error('requested channel id higher than ' +
+          'maximum (' + this.maxChannels_ + ')'));
+    }
+
     log.debug('%1: creating channel %2 with options: %3',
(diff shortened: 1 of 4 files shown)


src/webrtc/peerconnection.ts
```diff
@@ -428,6 +483,11 @@ export class PeerConnectionClass implements PeerConnection<signals.Message> {
       throw new Error('label cannot be an empty string');
     }
 
+    if (Object.keys(this.channels_).length >= this.maxChannels_) {
```
bemasc: I think this should probably also reject if `options && options.id >= this.maxChannels_`.

bemasc:
:+1: with one nit.  Beautiful!

src/webrtc/peerconnection.ts
trevj: Good idea. Done.

PR #280, opened 2015-09-08
Added encryption transformer to churn

blanu:
Added encryption transformer configuration block to churn-pipe
Rewrote churn transformer initialization to use string to constructor map instead of if-else block.
Added encryption transformer to churn constructor map.
Alphabetized imports as specified in the style guide.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/280)


trevj:
We can close this now I think!

PR #281, opened 2015-09-09
add a map of obfuscator names to class constructors, for easier config

trevj:
Based on:
https://github.com/uProxy/uproxy-lib/pull/280

Uncomment the lines in `churn.ts` and you can play with the AES traffic shaper. Works for me in Chrome in Docker and without, although there is some issue with Firefox support.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/281)


trevj:
Update: remove remaining FTE comments, as discussed.

bemasc:
:+1:

PR #283, opened 2015-09-10
SignalBatcher, v2

trevj:
When I tried to integrate the current `SignalBatcher` with uProxy, I realised it was going to be horribly inefficient: since uProxy wraps signalling channel messages in `social.PeerMessage` messages, we would have to _re-base64 encode_ the _already base64-encoded_ messages from `SignalBatcher`.

This was wiping out most of the gains from compression, so I've refactored `SignalBatcher` to handle arbitrary objects.

Batching (or "flattening") was the main casualty of this but I ran some tests and it turns out that with compression batching is virtually useless: while in one example batching reduced the raw size of the JSON from 2666 characters to 2173, with gzip compression it's a much tinier reduction from 729 to 711 bytes).

Tested with uProxy.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/283)

### DIFF
--- a/src/bridge/bridge.spec.ts
+++ b/src/bridge/bridge.spec.ts
@@ -47,6 +47,100 @@ var noMoreCandidatesSignal: peerconnection_types.Message = {
 // Static helpers.
 ////////
 
+describe('isTerminatingSignal', function() {
+  it('ignores messages without signals', () => {
+    expect(bridge.isTerminatingSignal({
+      errorOnLastMessage: true
+    })).toBeFalsy();
+  });
+
+  it('handles non-terminating PLAIN signal', () => {
+    expect(bridge.isTerminatingSignal({
+      signals: {
+        PLAIN: [
+          {
+            type: peerconnection_types.Type.CANDIDATE
+          }
+        ]
+      }
+    })).toBeFalsy();
+  });
+
+  it('handles terminating PLAIN signal', () => {
+    expect(bridge.isTerminatingSignal({
+      signals: {
+        PLAIN: [
+          {
+            type: peerconnection_types.Type.NO_MORE_CANDIDATES
+          }
+        ]
+      }
+    })).toBeTruthy();
+  });
+
+  it('handles non-terminating CHURN signal', () => {
+    expect(bridge.isTerminatingSignal({
+      signals: {
+        CHURN: [
+          {
+            caesar: 94
+          }
+        ]
+      }
+    })).toBeFalsy();
+  });
+
+  it('handles terminating CHURN signal', () => {
+    expect(bridge.isTerminatingSignal({
+      signals: {
+        CHURN: [
+          {
+            webrtcMessage: {
+              type: peerconnection_types.Type.NO_MORE_CANDIDATES
+            }
+          }
+        ]
+      }
+    })).toBeTruthy();
+  });
+
+  it('rejects multiple providers', () => {
+    expect(() => {
+      bridge.isTerminatingSignal({
+        signals: {
+          CHURN: [
+            {
+              caesar: 94
+            }
+          ],
+          HOLO_ICE: [
+            {
+              caesar: 94
+            }
+          ]
+        }
+      });
+    }).toThrow();
+  });
+
+  it('rejects multiple signals', () => {
+    expect(() => {
+      bridge.isTerminatingSignal({
+        signals: {
+          CHURN: [
+            {
+              caesar: 94
+            },
+            {
+              caesar: 94
+            }
+          ]
+        }
+      });
+    }).toThrow();
+  });
+});
+
 describe("makeSingleProviderMessage", function() {
   it('basic', () => {
     var signals = [
--- a/src/bridge/bridge.ts
+++ b/src/bridge/bridge.ts
@@ -81,16 +81,34 @@ export interface SignallingMessage {
 // Returns true iff message is a "terminating" message,
 // i.e. NO_MORE_CANDIDATES. This is intended for use in copy/paste
 // scenarios where it's important for the caller to know when
-// signalling is complete.
+// signalling is complete. Handles all providers known to the bridge
+// but does *not* handle "flattened" signalling messages, nor
+// messages with multiple providers.
 export var isTerminatingSignal = (message:SignallingMessage) : boolean => {
-  if (ProviderType[ProviderType.CHURN] in message.signals ||
-      ProviderType[ProviderType.HOLO_ICE] in message.signals) {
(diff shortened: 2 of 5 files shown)


bemasc:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[src/bridge/bridge.ts, line 85 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283#-JyslYH0Y2Hh4zH3Xi1t)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1cd02808e9ac20ad0a5af3d7227e6ef8d8330d40/src/bridge/bridge.ts#L85)):</sup>
What about the unobfuscated case?

---

<sup>**[src/copypaste-socks/freedom-module.ts, line 60 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283#-JysmeCPl1t2l63G4YB9)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1cd02808e9ac20ad0a5af3d7227e6ef8d8330d40/src/copypaste-socks/freedom-module.ts#L60)):</sup>
I'd prefer `(signal) => { parentModule.emit('signalForPeer', signal); },`.  Generally, I think that `bind` is so terrible in Typescript that it should be regarded as deprecated.

---

<sup>**[src/copypaste-socks/freedom-module.ts, line 117 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283#-JyssXf-F5PTYg2gHH0C)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1cd02808e9ac20ad0a5af3d7227e6ef8d8330d40/src/copypaste-socks/freedom-module.ts#L117)):</sup>
`encodedMessage` could be malformed and fail to decode.  Probably deserves some sort of error handling.

---

<sup>**[src/copypaste-socks/freedom-module.ts, line 120 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283#-JyssL_Xe_bCIJHKTQmk)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1cd02808e9ac20ad0a5af3d7227e6ef8d8330d40/src/copypaste-socks/freedom-module.ts#L120)):</sup>
`messages` is untrusted here.  Calling `forEach` should be safe, but could certainly throw an exception if the message is malformed.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283)

jpevarnek:
A couple questions for clarification in addition to what Ben had, but I think it looks pretty good!

---

Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

<sup>**[src/bridge/onetime.ts, line 23 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283#-JytOHlzYt_RZo6vp-yF)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1cd02808e9ac20ad0a5af3d7227e6ef8d8330d40/src/bridge/onetime.ts#L23)):</sup>
Can you clarify that messages that get evaluated as `isTerminating_` do not actually get included in the emitted batch (or anywhere?) (or if that is wrong, I'm reasonably sure I am incorrectly parsing the code)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283)

trevj:
Thanks everyone!

---

Review status: 1 of 5 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

<sup>**[src/bridge/bridge.ts, line 85 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283#-JyslYH0Y2Hh4zH3Xi1t)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1cd02808e9ac20ad0a5af3d7227e6ef8d8330d40/src/bridge/bridge.ts#L85)):</sup>
Good point! Done, plus tests.

---

<sup>**[src/bridge/onetime.ts, line 23 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283#-JytOHlzYt_RZo6vp-yF)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1cd02808e9ac20ad0a5af3d7227e6ef8d8330d40/src/bridge/onetime.ts#L23)):</sup>
Correct. Good idea. Done.

---

<sup>**[src/copypaste-socks/freedom-module.ts, line 60 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283#-JysmeCPl1t2l63G4YB9)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1cd02808e9ac20ad0a5af3d7227e6ef8d8330d40/src/copypaste-socks/freedom-module.ts#L60)):</sup>
At first I was loathe to add extra lines when I could use a feature of JavaScript I actually like...but then I looked into what a mess `bind` is in TypeScript...thanks :-) (also, sadness)

---

<sup>**[src/copypaste-socks/freedom-module.ts, line 117 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283#-JyssXf-F5PTYg2gHH0C)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1cd02808e9ac20ad0a5af3d7227e6ef8d8330d40/src/copypaste-socks/freedom-module.ts#L117)):</sup>
Good point, although the UI will only call this function once `validateSignalMessage` has succeeded. I added a comment to explain this.

---

<sup>**[src/copypaste-socks/freedom-module.ts, line 120 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283#-JyssL_Xe_bCIJHKTQmk)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/1cd02808e9ac20ad0a5af3d7227e6ef8d8330d40/src/copypaste-socks/freedom-module.ts#L120)):</sup>
Right. I just realised what a mess `handleSignalMessage` is: the return type in the `PeerConnection` interface is `void` while the concrete implementation returns a Promise, which can reject. I've filed an issue to clarify this:
https://github.com/uProxy/uproxy/issues/1885

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283)

bemasc:
:+1: 

---

Review status: 1 of 5 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/283)

PR #286, opened 2015-09-11
adventure should run with less logging, for performance purposes

trevj:
I just realised we run a bunch of throughput measurements with logging set to debug...that's probably not ideal. I don't think it makes much difference but why not (I performed a quick test with with `INFO` my throughput changed from 942Kb/sec to 985 which is most likely noise but who knows).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/286)


bemasc:
:+1:

PR #289, opened 2015-09-15
Run TCP integration tests with grunt-jasmine-firefoxaddon

agallant:
Passes locally, runs on Shippable but times out on `serves multiple clients` and `wrapper connectionsCount`. Will investigate/poke to see if it resolves, but wanted to open pull request for consideration. Thanks!

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/289)


.shippable.yml
```diff
@@ -18,4 +18,4 @@ install:
   - "firefox -v || echo 'No firefox installed.'"
   - "shippable_retry ./setup.sh install"
 script:
-  - "grunt test"
+  - "grunt test --firefox-bin /usr/bin/firefox"
```
trevj: Was this to workaround path lookup issues on Linux?

trevj:
Our Shippable setup has been busted for some time, unfortunately, so I'm not too worried about that. Many many thanks for sending this, it looks great to me. :+1:

trevj:
Related issue:
https://github.com/uProxy/uproxy/issues/840

.shippable.yml
soycode: Yeah exactly - jpm currently uses hardcoded default paths for Firefox that are actually incorrect on Debian/Ubuntu systems. See: https://github.com/canuckistani/fx-runner/pull/1

PR #291, opened 2015-09-17
configurable obfuscators, with plumbing through to the adventure app

trevj:
This got big, sorry.

In short, it's the plumbing required to configure the obfuscator, all the way from the adventure sample app through to churn pipe.

Summary of changes:
 * define `ObfuscatorConfig`, which churn accepts as an optional argument
 * move all the sample config stuff out of churn into the relevant obfuscator source files
 * add an adventure command which configures the obfuscator

The end result of all this is that it's pretty easy to write a script to benchmark our obfuscators:
obfuscator benchmarks

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/291)


trevj:
@blanu this is probably of most interest to you.

trevj:
BTW, it's clear we need to redesign the `Transformer` interface but I didn't want to tackle that right now. Removing `setKey` is the obvious first choice.

src/churn-pipe/churn-pipe.ts
```diff
@@ -157,14 +135,21 @@ class Pipe {
...
+      if (!(config.name in transformers)) {
+        throw new Error('unknown transformer: ' + config.name);
+      }
+
+      log.info('%1: using %2 obfuscator', this.name_, config.name);
+      this.transformer_ = new transformers[config.name]();
+      if (config.config !== undefined) {
+        this.transformer_.configure(config.config);
+      } else {
+        log.warn('%1: no config specified for obfuscator, using default', this.name_);
```
bemasc: Strictly speaking, if someone called `setTransformer` twice, with no config the second time, it would log this message but it wouldn't be using the default.

src/churn-pipe/churn-pipe.ts
```diff
@@ -157,14 +135,21 @@ class Pipe {
...
-      log.info('%1: using %2 transformer', this.name_, transformerName);
-      this.transformer_ = makeTransformer_(transformerName, key, config);
+      if (!(config.name in transformers)) {
+        throw new Error('unknown transformer: ' + config.name);
+      }
+
+      log.info('%1: using %2 obfuscator', this.name_, config.name);
+      this.transformer_ = new transformers[config.name]();
+      if (config.config !== undefined) {
+        this.transformer_.configure(config.config);
```
bemasc: `config.config` seems like something work avoiding.  Let's call one of these something else, like "settings".

src/churn/churn.ts
```diff
@@ -103,13 +101,17 @@ export var filterCandidatesFromSdp = (sdp:string) : string => {
   };
 
   // Generates a key suitable for use with CaesarCipher, viz. 1-255.
-  var generateCaesarKey_ = (): number => {
+  var generateCaesarKey_ = (): caesar.Config => {
```
bemasc: How about calling it `generateCaesarConfig_`?

src/churn-pipe/churn-pipe.ts
trevj: Right. Let's say...this code is not exactly where I'd like it to be. An obvious improvement is to pass the `ObfuscatorConfig` to the constructor. I was punting on that but am happy to do it in this pull request.

src/churn/churn.ts
trevj: Good catch. Done.

src/churn-pipe/churn-pipe.ts
bemasc: I was just suggesting changing the message :).  It could use the name of the current transformer, or it could just say "current transformer".

src/churn/churn.types.ts
```diff
@@ -8,4 +8,12 @@ export interface ChurnSignallingMessage {
   webrtcMessage ?:signals.Message;
   publicEndpoint ?:net.Endpoint;
   caesar ?:number;
+  obfuscator ?:ObfuscatorConfig;
+}
+
+export interface ObfuscatorConfig {
```
bemasc: This creates a naming conflict: sometimes it's a transformer, and sometimes it's an obfuscator.  Let's pick one.

blanu: I am in favor of phasing out terms relating to obfuscation in favor of something like transformers, as per our previous discussion on how obfuscation can have negative connotations.

blanu:
Reviewed 13 of 14 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

<sup>**[src/adventure/freedom-module.ts, line 74 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVchLMTBumfG5Yq_VE)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/adventure/freedom-module.ts#L74)):</sup>
I think transformerName and transformerConfig would be better names.

---

<sup>**[src/adventure/freedom-module.ts, line 94 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVdRBQcEQG18JDZNit)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/adventure/freedom-module.ts#L94)):</sup>
I think a good command name here would be "shapeshift"

---

<sup>**[src/adventure/freedom-module.ts, line 104 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVe7Ebuu-c3IfxUETx)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/adventure/freedom-module.ts#L104)):</sup>
Why 8?

---

<sup>**[src/adventure/freedom-module.ts, line 107 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVdcXrMCh5leRUv8pH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/adventure/freedom-module.ts#L107)):</sup>
It would be great to have a comment with an example command (or a few examples). Then it would be easier to see if the implemented parser is correct for the proposed examples.

---

<sup>**[src/adventure/freedom-module.ts, line 145 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVeuDUVIwlQN2zz_tn)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/adventure/freedom-module.ts#L145)):</sup>
This is a matter of preference, but if you have a parameter that is often going to be set to undefined, you can move it to the end of the argument list, at which point it can be elided. So for instance, if you switched the third and fourth arguments then "bridge.best('sockstortc', pcConfig, undefined, obfuscatorConfig)" could be written "bridge.best('sockstortc', pcConfig, obfuscatorConfig)".

---

<sup>**[src/fancy-transformers/byteSequenceShaper.ts, line 52 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVg_LzkBhrZjg1xOqZ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/fancy-transformers/byteSequenceShaper.ts#L52)):</sup>
All zeros is not going to be very helpful for debugging. It will be hard to see in the output as many protocols already begin with all zeros. Also if using Wireshark for testing it displays all non-ASCII characters as "." unless you switch to hex mode. So the best thing to do for testing is to use an all-ASCII byte sequence.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291)

src/churn-pipe/churn-pipe.ts
trevj: Oh :-) I changed it to `no transformer config specified` which hopefully will stand out in logs should we ever forget to specify a random config...thanks!

src/churn/churn.types.ts
trevj: Good points! I have changed to "transformer" throughout.

trevj:
Thanks for the review!

---

Review status: 7 of 14 files reviewed at latest revision, 10 unresolved discussions, some commit checks pending.

---

<sup>**[src/adventure/freedom-module.ts, line 74 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVchLMTBumfG5Yq_VE)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/adventure/freedom-module.ts#L74)):</sup>
Done!

---

<sup>**[src/adventure/freedom-module.ts, line 94 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVdRBQcEQG18JDZNit)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/adventure/freedom-module.ts#L94)):</sup>
I've changed it to "transform".

---

<sup>**[src/adventure/freedom-module.ts, line 104 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVe7Ebuu-c3IfxUETx)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/adventure/freedom-module.ts#L104)):</sup>
That's the length of ` config `. I changed the code a little to make this clearer.

---

<sup>**[src/adventure/freedom-module.ts, line 107 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVdcXrMCh5leRUv8pH)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/adventure/freedom-module.ts#L107)):</sup>
Totally agree. Done.

---

<sup>**[src/adventure/freedom-module.ts, line 145 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVeuDUVIwlQN2zz_tn)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/adventure/freedom-module.ts#L145)):</sup>
Totally agree. In this case, however, I didn't want to change the interface. This way, we don't have to change the uProxy frontend (which won't be ready to send a transformer config for some time).

---

<sup>**[src/fancy-transformers/byteSequenceShaper.ts, line 52 \[r2\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291#-JzVg_LzkBhrZjg1xOqZ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/0b278940a19ef37edc70a9f57297c4de2b82882f/src/fancy-transformers/byteSequenceShaper.ts#L52)):</sup>
OK. I changed it back to `OH HELLO`. The only reason I changed it was I was having other issues getting some of the transformers to run, and it turns out this isn't the culprit. Thanks!

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291)

src/churn-pipe/churn-pipe.ts
trevj: Hmm :-) I renamed it `transformerConfig`, what do you think?

blanu:
:+1: 

---

Reviewed 4 of 6 files at r3, 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/291)

src/churn-pipe/churn-pipe.ts
bemasc: :+1:

PR #292, opened 2015-09-18
Fixed a bug in byte sequence shaper

blanu:
Fixed a bug in restoring where matching would fail if the offset was not 0 or the packet length was longer than the sequence length.
Updated some type formatting.
Changed var to let.
Fixed one scope error detected by changing var to let.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/292)

### DIFF
--- a/src/fancy-transformers/byteSequenceShaper.ts
+++ b/src/fancy-transformers/byteSequenceShaper.ts
@@ -4,7 +4,7 @@ import arraybuffers = require('../arraybuffers/arraybuffers');
 import logging = require('../logging/logging');
 import random = require('../crypto/random');
 
-let log :logging.Log = new logging.Log('fancy-transformers');
+const log :logging.Log = new logging.Log('fancy-transformers');
 
 // Configuration where the sequences have been encoded as strings.
 // This is the interface that configure() expects as an argument.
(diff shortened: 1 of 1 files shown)


trevj:
:+1:

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[src/fancy-transformers/byteSequenceShaper.ts, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/292#-JzkH5hkPCnROTehdVRi)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/61c14f59a5d11280a57e9f82a69b6993d161424e/src/fancy-transformers/byteSequenceShaper.ts#L7)):</sup>
Do you think this would be a good candidate for `const`?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/292)

blanu:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[src/fancy-transformers/byteSequenceShaper.ts, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/292#-JzkH5hkPCnROTehdVRi)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/61c14f59a5d11280a57e9f82a69b6993d161424e/src/fancy-transformers/byteSequenceShaper.ts#L7)):</sup>
Okay I have made this change.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/292)

PR #293, opened 2015-09-18
Fixed a bug in the command parser

blanu:
The command parser had a bug where if two commands came in a single TCP read, the parser would break.
This has been changed so that the parser buffer however much data it gets and then splits on newline.
This eliminates the need for a hack in the command generator where a half second sleep is inserted between in command.
I have not tested this code yet as I do not know how to run the adventure sample app.
This is more of an exploratory pull request to bring up the idea of fixing the parser.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/293)

### DIFF
--- a/src/adventure/freedom-module.ts
+++ b/src/adventure/freedom-module.ts
@@ -76,15 +76,18 @@ function serveConnection(connection :tcp.Connection) :void {
       arraybuffers.stringToArrayBuffer('\n')
     ));
     if (index == -1) {
+      // A whole command has not been received yet. Continue receiving data.
       connection.dataFromSocketQueue.setSyncNextHandler(processCommands);
     } else {
+      // At least one whole command has been received. Attempt to parse it.
       let parts = arraybuffers.split(recvBuffer, index);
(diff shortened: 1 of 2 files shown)


trevj:
Thanks for looking at this! I just tried it out locally (by merging this into my `trevj-obfuscator-config` branch) and I'm afraid I still need the sleep in my Python script -- that's pretty weird, since it seems to work fine if I telnet to port 9000.

BTW, you need to click Launch on the chrome://extensions page in order to start adventure in Chrome.

blanu:
Try it now!

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/293)

trevj:
Thanks for fixing this! I just merged recent changes and verified this pull request needs sleep no more:
https://github.com/uProxy/uproxy-docker/pull/23

:+1:

PR #295, opened 2015-09-22
Chat app refactor, with a Firefox version

trevj:
Related issues:
- https://github.com/uProxy/uproxy/issues/1840
- https://github.com/uProxy/uproxy/issues/419

This refactors our chat apps a little, merging `simple-churn-chat-chromeapp` and `simple-freedom-chat` into `simple-chat` (with Chrome, web, and - finally! - Firefox variants).

This means there's no longer a churn _specific_ app -- however, I think this is okay because we generally only use the chat app when debugging low level peerconnection stuff, e.g. Firefox 42+ breakages. I've added a comment to the source explaining how to enable obfuscation.

I'm more worried that the pattern I've followed, first used for copypaste SOCKS, is kinda bad: it duplicates the HTML page and `glue.ts`. However, I'm happy that we're reducing the number of sample apps and duplicated TypeScript code.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/295)


bemasc:
:+1:

PR #297, opened 2015-09-24
some simple simple socks cleanup

trevj:
Main thing is to make it easy to use another transformer (line 48). Also a few bits and bobs.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/297)


trevj:
@blanu, do you think this would have helped you with testing the obfuscators? There's just one line you have to change now.

bemasc:
:+1:

PR #299, opened 2015-09-28
Fixed arithmetic coder tests

blanu:
Arithmetic coder tests were failing and now they pass

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/299)

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -203,6 +203,12 @@ taskManager.add 'unit_test', [
   'jasmine:queue'
 ]
 
+taskManager.add 'custom-test', [
+  'base'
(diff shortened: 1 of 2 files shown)


Gruntfile.coffee
```diff
@@ -203,6 +203,12 @@ taskManager.add 'unit_test', [
   'jasmine:queue'
 ]
 
+taskManager.add 'custom-test', [
+  'base'
+  'browserify:arithmeticSpec'
+  'jasmine:fancyTransformers'
```
trevj: why not just add `jasmine:fancyTransformers` to `unit_test`?

PR #300, opened 2015-09-28
Add a function to construct all tests

jpevarnek:
Well, as I said in the main uproxy repo, I'm sorry that this function is
so ugly.  Grunt sucks, and I am totally unable to think of a better way
to do this...I'd appreciate any suggestions though :).

TODO: move to gulp.

BTW, I do get to simplify this somewhat if I always assume that the rootDir is 'src' (true for uproxy and uproxy-lib).  Thoughts on that?

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/300)

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -1,9 +1,448 @@
+fs = require('fs')
+path = require 'path'
+rules = require './build/tools/common-grunt-rules'
 TaskManager = require './build/tools/taskmanager'
+
+#-------------------------------------------------------------------------
+devBuildPath = 'build/dev/uproxy-lib'
+thirdPartyBuildPath = 'build/third_party'
+localLibsDestPath = 'uproxy-lib'
+Rule = new rules.Rule({
+  # The path where code in this repository should be built in.
+  devBuildPath: devBuildPath,
+  # The path from where third party libraries should be copied. e.g. as used by
+  # sample apps.
+  thirdPartyBuildPath: thirdPartyBuildPath,
+  # The path to copy modules from this repository into. e.g. as used by sample
+  # apps.
+  localLibsDestPath: localLibsDestPath
+})
+
+readJSONFile = (file) -> JSON.parse(fs.readFileSync(file, 'utf8'))
+
+browserifyIntegrationTest = (path) ->
+  Rule.browserifySpec(path, {
+    browserifyOptions: { standalone: 'browserified_exports' }
+  })
+
+gruntConfig =
+  pkg: readJSONFile('package.json')
+
+  copy:
+    # Copy all src files into the directory for compiling and building.
+    src:
+      files: [
+        {
+            nonull: true,
+            expand: true,
+            cwd: 'src/',
+            src: ['**/*'],
+            dest: devBuildPath,
+            onlyIf: 'modified'
+        }
+      ]
+    # Copy releveant files to distribution directory.
+    dist:
+      files: [
+        {
+            nonull: true,
+            expand: true,
+            cwd: devBuildPath,
+            src: ['**/*',
+                  '!**/*.spec.js',
+                  '!**/*.ts',
+                  '**/*.d.ts'],
+            dest: 'build/dist/',
+            onlyIf: 'modified'
+        }
+      ]
+
+    # Copy the freedom output file to sample apps
+    # Rule.copyLibs [npmModules], [localDirectories], [thirdPartyDirectories]
+    libsForSimpleFreedomChat:
+      Rule.copyLibs
+        npmLibNames: ['freedom']
+        pathsFromDevBuild: ['loggingprovider']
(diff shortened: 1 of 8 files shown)


jpevarnek:
Diff for if we can assume everything is always under src (note: not tested, may have stupid bugs).  I think it is easier to understand what's going on in this version, and easier to deal with using it, but I don't LOVE the idea of loosing that level of genericism.

```
diff --git a/src/build-tools/common-grunt-rules.ts b/src/build-tools/common-grunt-rules.ts
index 29d3032..792a500 100644
--- a/src/build-tools/common-grunt-rules.ts
+++ b/src/build-tools/common-grunt-rules.ts
@@ -223,20 +223,20 @@ export class Rule {
    * rootDir is the directory under which the layout will match what this file
    * expects for paths being passed in (i.e. under devBuildPath).
    *
-   * getTests('src', gruntConfig);
+   * getTests('', gruntConfig);
    *   This adds tests for all the directories under src/ and browserifies all
    *   spec files within those directories.
-   * getTests('src', gruntConfig, 'generic_ui/scripts');
+   * getTests('generic_ui/scripts', gruntConfig);
    *   This adds tests for all the directories under src/generic_ui/scripts and
    *   browserifies spec files within those directories.  All paths will be
    *   relative to src/.
-   * getTests('src', gruntConfig, undefined, ['integration-tests']);
+   * getTests('', gruntConfig, ['integration-tests']);
    *   This adds tests for all the directories under src/ and browserifies all
    *   spec files within those directories.  Any directories named
    *   "integration-tests" will not be examined.
    */
-  public getTests = (rootDir :string, gruntConfig :{[c :string] :{[t :string] :Object}},
-                     current?:string, ignore?:string[], coverage?:boolean) => {
+  public getTests = (dir :string, gruntConfig :{[c :string] :{[t :string] :Object}},
+                     ignore?:string[], coverage?:boolean) => {
     if (typeof ignore === 'undefined') {
       ignore = [];
     }
@@ -245,26 +245,22 @@ export class Rule {
       coverage = false;
     }
 
-    if (typeof current === 'undefined') {
-      current = '';
-    }
-
-    var dir = path.join(rootDir, current);
+    var absoluteDir = path.join('src', dir);
 
     var tasks :string[] = [];
     var childTasks :string[] = [];
 
-    var files = fs.readdirSync(dir);
+    var files = fs.readdirSync(absoluteDir);
     for (var f in files) {
       var file = files[f];
       if (ignore.indexOf(file) !== -1) {
         continue;
       }
 
-      var stats = fs.statSync(path.join(dir, file));
+      var stats = fs.statSync(path.join(absoluteDir, file));
       if (stats.isDirectory()) {
         childTasks = childTasks.concat(
-            this.getTests(rootDir, gruntConfig, path.join(current, file), ignore, coverage));
+            this.getTests(dir, gruntConfig, file), ignore, coverage));
         continue;
       }
 
@@ -273,7 +269,7 @@ export class Rule {
         continue;
       }
 
-      var loc = path.join(current, match[1]);
+      var loc = path.join(dir, match[1]);
       var browserifyName = loc + 'Spec';
       var browserifyRule = this.browserifySpec(loc);
 
@@ -287,8 +283,8 @@ export class Rule {
     }
 
     if (tasks.length) {
-      var testName = current;
-      var testRule = this.jasmineSpec(current);
+      var testName = dir;
+      var testRule = this.jasmineSpec(dir);
 
       if (coverage) {
         testName += 'Cov';
```

trevj:
Thanks for looking at this!

I agree `getTest`'s args are complex. I think you'd have more luck abstracting it more. Essentially, you want a simple function that returns (as an array of strings) `src/*/*.spec.js`:

Then you could write something like:
```coffeescript
taskManager.add 'browserifySpecs', [
  'base'
].concat(Rule.browserifySpec spec for spec in Rule.getTests())

taskManager.add 'unit_test', [
  'browserifySpecs'
].concat(Rule.jasmineSpec spec for spec in Rule.getTests())
```

Make sense?

jpevarnek:
That makes sense, I personally like the more abstract idea of being able to say "I want to build and run a test for ___" and being able to then just do a single function call to set all of that up.

trevj:
I think it would be clearer again if we combined Rule.browserifySpec and Rule.jasmineSpec, then we could have:
```coffeescript
taskManager.add 'unit_test', [
  'base'
].concat(Rule.buildAndRunSpec spec for spec in Rule.getTests())
```

trevj:
Related issue:
https://github.com/uProxy/uproxy/issues/1741

trevj:
So...I'm hoping to make master the default branch soon. If you've no objections, I'm going to close this -- please re-open on `master` when it's ready, would love to have this in.

jpevarnek:
Yup, that's fine.  Sorry about the delay here.

trevj:
No worries, hope to see this again on the other side.

PR #306, opened 2015-10-19
rename adventure to zork

trevj:
Shorter, catchier, more memorable.

I wanted to name it this originally (hence the `xyzzy` easter egg) but chickened out...as it's become more and more important, I just think "zork" will be easier for people to remember what the hell we're talking about.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/306)


bemasc:
:+1:

PR #308, opened 2015-10-23
zork-node

ryscheng:
Done-ish!

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/308)

### DIFF
--- a/src/samples/zork-node/index.js
+++ b/src/samples/zork-node/index.js
@@ -1,10 +1,8 @@
 var freedom = require("freedom-for-node");
-console.log('Loading sample app');
 freedom.freedom('./uproxy-lib/zork/freedom-module.json', {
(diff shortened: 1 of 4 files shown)


src/samples/zork-node/index.js
```diff
@@ -0,0 +1,11 @@
+var freedom = require("freedom-for-node");
+console.log('Loading sample app');
```
trevj: I think we can remove the `console.log` statements (`console.error` can stay).

trevj:
Super cool. I had no idea `freedom-for-node` was so far along -- WebRTC excepted :-) We should add a note to the README (in particular, explain that it *does not work yet!*)...can you add a one-liner explaining how to run? Then, :+1: and huge thanks for doing this.

ryscheng:
Done

PR #310, opened 2015-10-26
add a social provider for proxying with zork instances

trevj:
First cut of a social provider which enables proxying with Zork instances. It's a social2 provider in which each instance is exposed as a friend in the contact list, and instances can be added via the invite flow.

An SSH tunnel is established with the server, across which signalling messages are exchanged. Currently, the server is expected to have a `giver` account having a password `giver`. This will change to key-based authentication, with the keys emailed to users.

Lots of polish to come.

In-progress uProxy pull request:
https://github.com/uProxy/uproxy/compare/trevj-social-cloud

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/310)

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -12,6 +12,7 @@ taskManager.add 'base', [
   'ts:srcInCoreEnv'
   'browserify:loggingProvider'
   'browserify:churnPipeFreedomModule'
+  'browserify:cloudSocialProviderFreedomModule'
 ]
 
 taskManager.add 'samples', [
@@ -28,9 +29,6 @@ taskManager.add 'samples', [
 # Makes the distribution build.
 taskManager.add 'dist', [
   'base'
-  'samples'
-  'test'
-  'coverage'
   'copy:dist'
 ]
 
@@ -535,6 +533,17 @@ module.exports = (grunt) ->
       copypasteChatFreedomModule: Rule.browserify 'copypaste-chat/freedom-module'
       copypasteSocksFreedomModule: Rule.browserify 'copypaste-socks/freedom-module'
       echoServerFreedomModule: Rule.browserify 'echo/freedom-module'
+      cloudSocialProviderFreedomModule: Rule.browserify('cloud/social/freedom-module', {
+        alias : [
+          # Shims for node's dns and net modules from freedom-social-xmpp,
+          # with a couple of fixes.
+          './src/cloud/social/shim/net.js:net'
+          './src/cloud/social/shim/dns.js:dns'
+          # Subset of ssh2-streams (all except SFTP) which works well in
+          # the browser.
+          './src/cloud/social/alias/ssh2-streams.js:ssh2-streams'
+        ]
+      })
       simpleChatFreedomModule: Rule.browserify 'simple-chat/freedom-module'
       simpleSocksFreedomModule: Rule.browserify 'simple-socks/freedom-module'
       simpleTurnFreedomModule: Rule.browserify 'simple-turn/freedom-module'
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -29,6 +29,9 @@ taskManager.add 'samples', [
 # Makes the distribution build.
 taskManager.add 'dist', [
   'base'
+  'samples'
+  'test'
+  'coverage'
   'copy:dist'
 ]
 
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "jasmine-core": "^2.3.4",
     "request": "^2.53.0",
     "socks5-http-client": "^0.1.6",
+    "ssh2": "^0.4.11",
     "tsd": "^0.6.5",
     "typescript": "^1.6.2",
     "yargs": "^3.0.4"
--- a/src/cloud/social/alias/ssh2-streams.js
+++ b/src/cloud/social/alias/ssh2-streams.js
@@ -1,0 +1,11 @@
+// This is identical to ssh2-streams' index.js except for SFTPStream.
+// We do this because sftp.js contains a call to process.bindings
+// that raises an error in the browser. This is safe because we don't
+// actually care about using SFTP, it's just that ssh2's Client sets
+// some constants to point to some of SFTPStream's constants.
+module.exports = {
+  SFTPStream: {},
+  SSH2Stream: require('ssh2/node_modules/ssh2-streams/lib/ssh'),
+  utils: require('ssh2/node_modules/ssh2-streams/lib/utils'),
+  constants: require('ssh2/node_modules/ssh2-streams/lib/constants')
+};
--- a/src/cloud/social/alias/ssh2-streams.js
+++ b/src/cloud/social/alias/ssh2-streams.js
@@ -5,7 +5,7 @@
 // some constants to point to some of SFTPStream's constants.
 module.exports = {
   SFTPStream: {},
-  SSH2Stream: require('ssh2/node_modules/ssh2-streams/lib/ssh'),
-  utils: require('ssh2/node_modules/ssh2-streams/lib/utils'),
-  constants: require('ssh2/node_modules/ssh2-streams/lib/constants')
+  SSH2Stream: require('ssh2-streams/lib/ssh'),
+  utils: require('ssh2-streams/lib/utils'),
+  constants: require('ssh2-streams/lib/constants')
 };
--- a/src/cloud/social/freedom-module.json
+++ b/src/cloud/social/freedom-module.json
@@ -1,0 +1,125 @@
+{
+  "name": "Cloud Social Provider",
+  "description": "Exposes Zork instances as friends.",
+  "app": {
+    "script": [
+      "freedom-module.static.js"
+    ]
+  },
+  "default": "social2",
+  "provides": [
+    "social2"
+  ],
+  "api": {
+    "social2": {
+      "login": {
+        "type": "method",
+        "value": [{
+          "agent": "string",
+          "version": "string",
+          "url": "string",
+          "interactive": "boolean",
+          "rememberLogin": "boolean"
+        }],
+        "ret": {
+          "userId": "string",
+          "clientId": "string",
+          "status": "string",
+          "timestamp": "number"
+        }
+      },
+
+      "sendMessage": {
+        "type": "method",
+        "value": ["string", "string"]
+      },
+
+      "clearCachedCredentials": {
+        "type": "method",
+        "value": []
+      },
+
+      "getUsers": {
+        "type": "method",
+        "value": [],
+        "ret": "object"
+      },
+
+      "getClients": {
+        "type": "method",
+        "value": [],
+        "ret": "object"
+      },
+
+      "logout": {
+        "type": "method",
+        "value": []
+      },
+
+      "inviteUser": {
+        "type": "method",
+        "value": ["string"],
+        "ret": "Object"
+      },
+
+      "acceptUserInvitation": {
+        "type": "method",
+        "value": ["string"]
+      },
+
+      "blockUser": {
+       "type": "method",
+        "value": [{
+          "userId": "string"
+        }]
+      },
+
+      "removeUser": {
+       "type": "method",
+        "value": [ "string"]
+      },
+
+      "onMessage": {
+        "type": "event",
+        "value": {
+          "from": {
+            "userId": "string",
+            "clientId": "string",
+            "status": "string",
+            "lastUpdated": "number",
+            "lastSeen": "number"
+          },
+          "message": "string"
+        }
+      },
+
+      "onUserProfile": {
+        "type": "event",
+        "value": {
+          "userId": "string",
+          "lastUpdated": "number",
+          "name": "string",
+          "url": "string",
+          "imageData": "string",
+          "status": "number"
+        }
+      },
+
+      "onClientState": {
+        "type": "event",
+        "value": {
+          "userId": "string",
+          "clientId": "string",
+          "status": "string",
+          "lastUpdated": "number",
+          "lastSeen": "number"
+        }
+      }
+    }
+  },
+  "permissions": [
+    "core.tcpsocket",
+    "core.udpsocket",
+    "core.storage"
+  ]
+}
--- a/src/cloud/social/freedom-module.ts
+++ b/src/cloud/social/freedom-module.ts
@@ -1,0 +1,7 @@
+/// <reference path='../../../../third_party/typings/freedom/freedom-module-env.d.ts' />
+
+import provider = require('./provider');
+
+if (typeof freedom !== 'undefined') {
+  freedom().providePromises(provider.CloudSocialProvider);
+}
--- a/src/cloud/social/provider.ts
+++ b/src/cloud/social/provider.ts
@@ -18,10 +18,11 @@ const SSH_SERVER_PASSWORD = 'giver';
 const ZORK_HOST = 'zork';
 const ZORK_PORT = 9000;
 
-// TODO: what is the namespace for storage keys?
+// Key under which our contacts are saved in storage.
 const STORAGE_KEY = 'cloud-social-contacts';
 
-// Placed in storage under STORAGE_KEY.
+// Type of the object placed, in serialised form, in storage
+// under STORAGE_KEY.
 interface SavedContacts {
   addresses?: string[];
 }
--- a/src/cloud/social/provider.ts
+++ b/src/cloud/social/provider.ts
@@ -1,0 +1,459 @@
+/// <reference path='../../../../third_party/typings/es6-promise/es6-promise.d.ts' />
+/// <reference path='../../../../third_party/typings/freedom/freedom-module-env.d.ts' />
+/// <reference path='../../../../third_party/typings/node/node.d.ts' />
+/// <reference path='../../../../third_party/typings/ssh2/ssh2.d.ts' />
+
+import logging = require('../../logging/logging');
+
+// https://github.com/borisyankov/DefinitelyTyped/blob/master/ssh2/ssh2-tests.ts
+import * as ssh2 from 'ssh2';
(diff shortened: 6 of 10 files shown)


trevj:
I forgot to mention: there's a bunch of dependencies here - mostly net-related shims for ssh2 - so you probably want to focus on this commit:
https://github.com/uProxy/uproxy-lib/commit/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0

dborkan:
Generally looks good!  Just some small stuff in this code review.

In the longer term I think there are questions to answer about where some of the instance message and SIGNAL_FROM_SERVER_PEER logic will live.  Just to document some of the offline discussion yesterday I think the options are:

1. On the server: this makes the most sense for keeping the uProxy and social provider code clean, but I worry that any logic on the server will be very hard to change once servers are deployed.
2. In the social provider (client): this will be easier to change than server logic, and it encapsulates some of the cloud-specific logic, but I worry if we make changes to format of instance messages or SIGNAL_FROM_SERVER_PEER behavior we are likely to forget about changing it in the cloud provider code.
3. In uProxy: this would mean special-case code for the cloud provider, but would allow us to easily make changes to the instance message or SIGNAL_FROM_SERVER_PEER format.  I think the changes for this would be to not expect instance message from cloud peers, and to assume all onMessage events from the cloud provider are type=SIGNAL_FROM_SERVER_PEER


---

Review status: 0 of 10 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

<sup>**[src/cloud/social/provider.ts, line 21 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1aF1p7l1uWL9F11oPc)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L21)):</sup>
The freedom storage provider currently puts everything in the same scope.  But I think you should be fine using 'cloud-social-contacts' here for now


---

<sup>**[src/cloud/social/provider.ts, line 53 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1aPCgYNafPaYVT5xIN)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L53)):</sup>
We can leave this for now, but I think ideally instance messages wouldn't be in the social provider (otherwise if we make changes to instance messages we might forget about this case).  Maybe another solution would be to modify uProxy core code so that it doesn't expect any instance messages from this social provider (adding a new option to network-options.ts in uProxy)?  Let me know what you think, we can add a GitHub issue to clean it up later


---

<sup>**[src/cloud/social/provider.ts, line 128 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1aPwIHsxgP4wy7wvXA)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L128)):</sup>
I think we ought to emit an 'onClientState' here (with the same data returned by makeClientState).  That will also help us clean things up if we remove the makeInstanceMessage logic (uProxy could create a new RemoteInstance for this cloud server at that time, with instanceId=clientId)


---

<sup>**[src/cloud/social/provider.ts, line 138 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1aSSq9du_S6gcG0KWF)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L138)):</sup>
Just curious - is there a reason we can't re-use an existing connection?


---

<sup>**[src/cloud/social/provider.ts, line 152 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1eKfKYS_RTl4xJfhGw)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L152)):</sup>
This is fine for now but I think we might want to revisit this - it seems a little weird to have what is normally uProxy logic here.

Some other alternatives:
-  if we get rid of the instance message logic (suggested above), we could modified uProxy so that it knows that all 'onMessage' events from the cloud social provider are type=SIGNAL_FROM_SERVER_PEER
- we could also make the cloud server reply with these messages, but the downside of that approach is it might be very hard to change the behavior of cloud servers once deployed, in case we ever want to modify the message format

Maybe just add a TODO here?


---

<sup>**[src/cloud/social/provider.ts, line 174 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1aRWpwMbCBiSKnsTYn)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L174)):</sup>
maybe change the name of 'v' to be like 'storedString'?


---

<sup>**[src/cloud/social/provider.ts, line 212 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1eeUxhLcFzR-k2KWS4)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L212)):</sup>
NIT: it might be good to reset this.savedAddresses_ here.  Otherwise if we ever call .login, then .logout, then .login again, the 2nd login won't get an onUserProfile events generated.

....But in practice uProxy creates a new freedom social object everytime you login so cleaning up state isn't so important


---

<sup>**[src/cloud/social/provider.ts, line 214 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1egUa-cDum6eod5VM1)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L214)):</sup>
NIT: social providers should emit an 'onUserProfile' here with the profile for the logged in user (so that uProxy can display their name, image, and URL).  It's not that important for cloud because we don't yet know those things, but maybe we want a TODO or other comment here?

But uProxy fails gracefully if those things are missing (they had often been missing for GTalk XMPP)


---

<sup>**[src/cloud/social/provider.ts, line 266 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1eZ_49OVA8AxCYLGv7)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L266)):</sup>
getUsers should return all the known UserProfiles (the profile for the logged in user, plus everything that would be returned by loadContacts).  However uProxy doesn't actually use this, so if you want to just change this to return Promise.reject() that's fine too.


---

<sup>**[src/cloud/social/provider.ts, line 271 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1edi9e562rFqSDKdml)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L271)):</sup>
Same as getUsers - can you either return all clients or you can just reject?


---

<sup>**[src/cloud/social/provider.ts, line 276 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1edzA_kPsv8viY9Tl8)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L276)):</sup>
Do you want to close any outstanding connections here?


---

<sup>**[src/cloud/social/provider.ts, line 287 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1eW-DFVEqUE3yBVtCZ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L287)):</sup>
It's fine to resolve and not return anything.  We are currently using 1 inviteUser method which handles both URL style invites that return networkData (like Firebase, Quiver), and ones where we expect the invite to be accepted in some other way (Cloud, GitHub)


---

<sup>**[src/cloud/social/provider.ts, line 296 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1ednx3kXNDPvvYpjqX)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L296)):</sup>
I think for this and other unimplemented methods it's probably best to reject - uProxy shouldn't call them anyways, and if it does rejecting will be better for detecting an error


---

<sup>**[src/cloud/social/provider.ts, line 332 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1elKakp76mUDoSpA1T)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L332)):</sup>
I think send_ is a bit confusing as a variable name here.  I think messageReceivedHandler_ or just received_ is more accurate.  Using send_ makes me think that the message is going from the local machine to the server (similar to sendMessage)


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310)

trevj:
Thanks for the detailed review!

Agree that hacking social messages is the biggest question here. On those three options:
 1. We have to worry about backwards compatibility and slow updates anyway, so I actually don't see much extra work here in terms of maintainability.
 1. I think this is the least desirable of the three! Sadly, what we have after this pull request :-)
 1. This implies uProxy is happy to speak with something that *isn't uProxy*.

In the medium-term, I can imagine a different model for the social provider in which cloud instances appear as *clients*, associated with a user on some social network like Quiver. That's some way away, though, and will require a non-trivial refactoring of the social code.

Can you imagine another case for #3?


---

Review status: 0 of 10 files reviewed at latest revision, 14 unresolved discussions.

---

<sup>**[src/cloud/social/provider.ts, line 21 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1aF1p7l1uWL9F11oPc)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L21)):</sup>
Awesome, thanks. I removed the TODO.


---

<sup>**[src/cloud/social/provider.ts, line 53 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1aPCgYNafPaYVT5xIN)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L53)):</sup>
See top-level comment.


---

<sup>**[src/cloud/social/provider.ts, line 128 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1aPwIHsxgP4wy7wvXA)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L128)):</sup>
Done.


---

<sup>**[src/cloud/social/provider.ts, line 138 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1aSSq9du_S6gcG0KWF)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L138)):</sup>
Yes: once Zork receives a `get` or `give` command then all further input and output is considered to be a signalling message. So, basically, it's a one-time thing.

I just realised there's one exception: when you've just successfully invited a cloud instance, *that* connection could be re-used. I've added a TODO for that.

We could also make Zork less dumb but that's best left for the future :-)


---

<sup>**[src/cloud/social/provider.ts, line 152 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1eKfKYS_RTl4xJfhGw)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L152)):</sup>
Agree it's weird. See my top-level comment.


---

<sup>**[src/cloud/social/provider.ts, line 174 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1aRWpwMbCBiSKnsTYn)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L174)):</sup>
Done.


---

<sup>**[src/cloud/social/provider.ts, line 212 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1eeUxhLcFzR-k2KWS4)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L212)):</sup>
Nice catch! Now, `loadContacts_` resets that field.

Agree that while not a problem in practise, it's easy to get right.


---

<sup>**[src/cloud/social/provider.ts, line 214 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1egUa-cDum6eod5VM1)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L214)):</sup>
Great, thanks. I added a TODO to emit that once we have an image.


---

<sup>**[src/cloud/social/provider.ts, line 266 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1eZ_49OVA8AxCYLGv7)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L266)):</sup>
Done. I think I had it this way because I wasn't always seeing the log line if I rejected (weird, I dunno why that would happen).


---

<sup>**[src/cloud/social/provider.ts, line 271 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1edi9e562rFqSDKdml)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L271)):</sup>
Done.


---

<sup>**[src/cloud/social/provider.ts, line 276 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1edzA_kPsv8viY9Tl8)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L276)):</sup>
Done.


---

<sup>**[src/cloud/social/provider.ts, line 287 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1eW-DFVEqUE3yBVtCZ)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L287)):</sup>
Oh great! Thanks.


---

<sup>**[src/cloud/social/provider.ts, line 296 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1ednx3kXNDPvvYpjqX)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L296)):</sup>
Done.


---

<sup>**[src/cloud/social/provider.ts, line 332 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310#-K1elKakp76mUDoSpA1T)** ([raw file](https://github.com/uproxy/uproxy-lib/blob/ddd21c0ea1ea046e6007ea39dd10c2025f73d4e0/src/cloud/social/provider.ts#L332)):</sup>
Much better! Thanks.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310)

dborkan:
:+1: Good stuff!!!


---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310)

dborkan:
Also related to case 3 above, I'm not sure of other cases where uProxy would speak to something non-uProxy (maybe some "microProxy" on Android if that happens?), but I do think it's inevitable that we will treat contacts on the cloud social provider differently, e.g. we will immediately show the "start getting access", and hide them from the share tab.  So I don't think it would be that difficult to make a few more exceptions in regards to not expecting instance messages, and treating all onMessage events as type=SIGNAL_FROM_SERVER_PEER


---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310)

trevj:
OK. Let's talk soon about these messages. #3 doesn't sound like such a bad solution. Thanks again for the review!


---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/310)

PR #313, opened 2015-11-05
src/cloud/provision is a Digital Ocean droplet provision in a freedom.js module

ryscheng:
Includes a sample app (provision-chromeapp) to test the functionality

Provisioner will launch oauth flow, generate SSH keys, upload keys and config to Digital Ocean and wait for machine to start

Also includes a method to destroy a droplet

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/313)


src/provision/freedom-module.js
```diff
@@ -0,0 +1,351 @@
...
+          return Promise.resolve({
+            message: "Droplet already created with name=" + name,
+            droplet: resp.droplets[i]
+          });
+        }
+      }
+
+      return this._doRequest("POST", "droplets", JSON.stringify({
+        name: name,
+        region: "nyc3",
```
soycode: Maybe region should be a parameter?

src/provision/freedom-module.js
```diff
@@ -0,0 +1,351 @@
+//var XMLHttpRequest = require('freedom-xhr').corexhr;
```
willscott: leftover?

src/provision/freedom-module.js
```diff
@@ -0,0 +1,351 @@
...
+  });
+};
+
+/*
+ * Generates an RSA keypair using forge
+ * @return {Object.<String,String>} public and private SSH keys
+ */
+Provisioner.prototype._generateKeyPair = function() {
+  "use strict";
+  var pair = forge.pki.rsa.generateKeyPair({bits: 1024, e: 0x10001});  // TODO: use async
```
willscott: why 1024 bit? i thought that was considered on the verge of insecurity. can we do ecc yet?

src/provision/freedom-module.js
```diff
@@ -0,0 +1,351 @@
...
+  var publicKey = forge.ssh.publicKeyToOpenSSH(pair.publicKey, 'info@uproxy.org');
```
willscott: why is the comment needed? maybe better to leave empty?

src/provision/freedom-module.js
```diff
@@ -0,0 +1,351 @@
...
+
+    Promise.all([
+      storage.get("DigitalOcean-" + name + "-PublicKey"),
+      storage.get("DigitalOcean-" + name + "-PrivateKey")
+    ]).then(function(val) {
+      if (val[0] === null ||
+         val[1] === null) {
+        result = this._generateKeyPair();
+        storage.set("DigitalOcean-" + name + "-PublicKey", result.public);
+        storage.set("DigitalOcean-" + name + "-PrivateKey", result.private);
```
willscott: do we have a bug to be able to store at least the private key encrypted with a passphrase, so that the user needs to decrypt it at time of use?

src/provision/freedom-module.js
```diff
@@ -0,0 +1,351 @@
...
+    }.bind(this)).then(function(resp) {
+      //console.log(resp);
+      this._sendStatus("CLOUD_WAITING_VM");
+      this._waitDigitalOceanActions(resolve, reject);
+    // Wait for all in-progress actions to complete
+    }.bind(this)).catch(function(err) {
+      console.error("Error w/DigitalOcean: " + err);
+      this._sendStatus("CLOUD_FAILED");
+      reject({
+        errcode: "",
```
willscott: no error code?

src/provision/freedom-module.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "provision",
+  "description": "Provisions VMs on Digital Ocean",
+  "app": {
+    "script": [
+      "/forge-min/forge.min.js",
```
willscott: do scripts with an absolute path work everywhere? we've normally been using relative paths.

src/samples/provision-chromeapp/background.js
```diff
@@ -0,0 +1,37 @@
...
+  var script = document.createElement('script');
+  script.src = 'freedom-for-chrome/freedom-for-chrome.js';
+  document.head.appendChild(script);
+  script.onload = function() {
+    console.log('loading freedom!');
+    freedom('uproxy-lib/provision/freedom-module.json', {
+      'logger': 'uproxy-lib/loggingprovider/freedom-module.json',
+      'debug': 'debug'
+    }).then(function(moduleFactory) {
+      var module = moduleFactory();
```
willscott: module, i think, is becoming a reserved word in js. maybe pick a different symbol?

package.json
```diff
@@ -33,11 +33,13 @@
     "circular-json": "^0.1.6",
     "crypto": "^0.0.3",
     "es6-promise": "^2.0.0",
+    "forge-min": "^0.6.20",
```
trevj: I haven't come across forge before!

The cloud social provider in here has been using ssh2 to establish SSL tunnels -- and ssh2 can generate public/private key pairs, too (actually ssh2-streams, one of its dependencies). This is fine for now but we should consider using one or the other to reduce bloat (and, if we do, the social provider may want to use the `Provisioner` class directly to avoid another browserify step).

src/provision/freedom-module.js
trevj: +1 let's remove the commented code

src/provision/freedom-module.js
trevj: For sure. I see the two key parameters being region and size. But we can tackle that in a followup.

src/provision/freedom-module.json
trevj: +1 let's use a relative path

src/provision/freedom-module.js
```diff
@@ -0,0 +1,351 @@
...
+  "CLOUD_FAILED": "Failed to complete cloud operation",
+  "CLOUD_INIT_ADDKEY": "Starting to add SSH key to cloud account",
+  "CLOUD_DONE_ADDKEY": "Done adding SSH key to cloud account",
+  "CLOUD_INIT_VM": "Starting to provision VM",
+  "CLOUD_DONE_VM": "Done provisioning VMs",
+  "CLOUD_WAITING_VM": "Waiting on VM",
+};
+
+var REDIRECT_URIS = [
+  "https://kmnjcbmibpajfljogomikcdlmhgpnolg.chromiumapp.org"
```
trevj: What's the significance of the hostname here?

src/provision/freedom-module.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "provision",
+  "description": "Provisions VMs on Digital Ocean",
```
trevj: This is fine but FYI I've generally found it easier to create a freedomjs module *after* the fact, i.e. once we know we need it. As I said in another comment, there's quite a lot of dependencies here so we might end up drawing the module boundary at just the social provider layer, instead, for module size if nothing else. Anyway, not a big deal for now.

trevj:
Awesome. The sample app is especially handy.

I've added a few low-level comments but I have two more general comments:
 1. We try to write in TypeScript :-) Aside from type checking, you could remove all the `.bind(this)` calls *and* use some nice ES6 features such as `let` or `const`.
 1. I think `src/cloud/provision` would be a great location for this.

Anyway, I'm happy enough to make the two above changes later -- more concerned with the basic functionality right now. As I wrote in my email, I'm having some trouble authenticating the sample app as-is. Something to do with the OAuth config?

src/provision/freedom-module.js
ryscheng: Great idea! Filed an issue to remember to followup
https://github.com/uProxy/uproxy/issues/2049

src/provision/freedom-module.js
ryscheng: Done

src/provision/freedom-module.js
ryscheng: Ah good catch. This was copied from Colony, which I think lowered it because it was running in the main thread, which blocked everything else. Now that it's in a freedom.js module, we shouldn't worry about blocking the thread as much.

ryscheng: Done

src/provision/freedom-module.js
ryscheng: Done

src/provision/freedom-module.js
ryscheng: Good point. Issue filed here:
https://github.com/uProxy/uproxy/issues/2050

src/provision/freedom-module.json
ryscheng: Yes, I agree with you. I needed to wrap the code in a freedom.js module for the sample app in order to use core.oauth. Plus, the SSH keygen will block for seconds, so I found this useful for that reason too. If you want to just link this into another module, that works too

src/provision/freedom-module.js
ryscheng: https://developer.chrome.com/apps/identity

Right now the same app will use the built-in core.oauth provider in freedom-for-chrome, which uses chrome.identity. This is the format of redirect URIs that need to be registered in order for chrome.identity to properly get the token.

ryscheng: Of course this will change when we link into uproxy, we should follow similar conventions that we use for our social providers.

package.json
ryscheng: Sure! We're only using it for keygen, so pretty easy to replace. It was copied from Colony so I knew it worked. Lemme file an issue to replace.
https://github.com/uProxy/uproxy/issues/2051

src/samples/provision-chromeapp/background.js
ryscheng: Done

src/provision/freedom-module.json
ryscheng: Done

src/provision/freedom-module.js
ryscheng: Done

ryscheng:
Okay! addressed most comments above. The ones that I didn't I filed GitHub issues for. I also fixed the extension ID in the sample app, which I believe was causing the oAuth errors. Let me know if this doesn't work!

trevj:
Working for me now! Wonderful stuff. :+1:

PR #314, opened 2015-11-10
invite codes for the cloud social provider

trevj:
Still prototyping...this makes the cloud social provider accept invite codes. An invite code encodes a server's address and credentials and, when shared, grants access to the server. I guess at some point we could make these URLs?

This in-progress pull request makes the cloud setup script output an invite code when it's done:
https://github.com/uProxy/uproxy-docker/compare/trevj-social-cloud?expand=1

Tested on my workstation and on a Digital Ocean provider.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/314)

### DIFF
--- a/src/cloud/social/provider.ts
+++ b/src/cloud/social/provider.ts
@@ -12,8 +12,6 @@ var Client = require('ssh2').Client;
 var log: logging.Log = new logging.Log('cloud social');
 
 const SSH_SERVER_PORT = 5000;
-const SSH_SERVER_USERNAME = 'giver';
-const SSH_SERVER_PASSWORD = 'giver';
 
 const ZORK_HOST = 'zork';
 const ZORK_PORT = 9000;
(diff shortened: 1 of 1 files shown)


dborkan:
:+1: for this change, except I think we should move the code from inviteUser to acceptUserInvitation.  In most cases (all networks except GitHub), inviteUser doesn't take any args but returns an Object which is then encoded as part of uProxy's invite URL.  Then this same data is typically passed as JSON to acceptUserInvitation (basically what you have inviteUser doing here).

Also from a UX perspective, I think adding a server corresponds more to accepting an invite that you received (unless you are deploying a new one).  inviteUser seems more like you already own a server and want to permission friends to use it (so you'd want to generate an invite URL to send to your friend)


---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/314)

trevj:
As discussed, let's keep the `inviteUser` approach as-is until we have a better idea on how deploying and inviting users will work in the new invitation flow world.

Thanks!

PR #318, opened 2015-11-17
move all transformers under transformers/

trevj:
Another daily code health workout.

Trying to reduce the number of top-level directories and eliminate hyphens in them, too.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/318)


bemasc:
:+1:

PR #319, opened 2015-11-18
implement acceptUserInvitation rather than inviteUser

trevj:
Seems to work well with the evolving login/invite experience.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/319)


lucyhe:
:+1:

PR #321, opened 2015-11-18
deps sweep

trevj:
All seems to work!

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/321)


trevj:
FYI, `grunt-vulcanize` seems to be the only one we can't upgrade because we're not on Polymer 1.0. That seems like a larger job than I'm willing to tackle right now :-)

bemasc:
:+1:

PR #326, opened 2015-11-23
fix undefined variable exception when no saved contacts

trevj:
Bah. Only affects someone who had been using cloud before we changed the storage format...

Makes me want to move this into the uProxy repo more :-)

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/326)


lucyhe:
:+1:

PR #331, opened 2015-12-04
Translated cloud provision freedom module to TypeScript

jamesliu96:
Translated from JavaScript, some types are still not clear enough.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/331)


src/cloud/provision/freedom-module.ts
```diff
@@ -0,0 +1,362 @@
...
+interface KeyPair {
+  private: string;
+  public: string;
+}
+
+class Provisioner {
+  private dispatch: Function;
+  private state: any;
+
+  constructor(dispatchEvent: Function, state: any) {
```
trevj: We find the automatic constructor assignment stuff useful, i.e.:
```typescript
constructor(private dispatchEvent_: Function, private state_: any) {}
```

Saves quite a few lines.

src/cloud/provision/freedom-module.ts
```diff
@@ -0,0 +1,362 @@
...
+   * @return {Promise.<Object>} oAuth response from Digital Ocean
+   *  {
+   *    access_token: "..",
+   *    expires_in: "..",
+   *    state: "..",
+   *    token_type: ".."
+   *  }
+   */
+  private doOAuth_ = () : Promise<Object> => {
+    return new Promise((resolve: Function, reject: Function) => {
```
trevj: Elsewhere in this repo we have TypeScript infer the types of `resolve` and `reject`. This means you can rewrite this line as:
```typescript
return new Promise((F, R) => {
```

src/cloud/provision/freedom-module.ts
```diff
@@ -0,0 +1,362 @@
...
+   * Try to retrieve SSH keys from storage.
+   * If not found, generate new ones and store
+   * @param {String} name name of the key (usually same as name of VM later)
+   * @return {Promise.<Object>} the SSH keys retrieved
+   * {
+   *    public: "...",
+   *    private: "..."
+   * }
+   */
+  private getSshKey_ = (name: string) : Promise<Object> => {
```
trevj: I think the return type is `Promise<KeyPair>`, right?

src/cloud/provision/freedom-module.ts
```diff
@@ -0,0 +1,362 @@
...
+  
+  /**
+   * Properly configure Digital Ocean with a single droplet of name:name
+   * Assumes we already have oAuth token and  SSH key in this.state
+   * This method will use this.waitDigitalOceanActions_() to wait until all actions complete
+   * before resolving
+   * @param {String} name of droplet
+   * @return {Promise.<Object>} resolves on success, rejects on failure
+   */
+  private setupDigitalOcean_ = (name: string) : Promise<Object> => {
```
trevj: I think the return type here is `Promise<void>`.

src/cloud/provision/freedom-module.ts
```diff
@@ -0,0 +1,362 @@
...
+   * @param {String} name name of the key (usually same as name of VM later)
+   * @return {Promise.<Object>} the SSH keys retrieved
+   * {
+   *    public: "...",
+   *    private: "..."
+   * }
+   */
+  private getSshKey_ = (name: string) : Promise<Object> => {
+    var storage = freedom["core.storage"]();
+    return new Promise((resolve: Function, reject: Function) => {
```
trevj: Again, ```new Promise((F, R) => {``` should work here.

trevj:
Thank you so much for doing this!

Just a few comments and suggestions to make it even better and I think we'll be good to go!

jamesliu96:
@trevj Sure. It's my first time writing in typescript and also for freedom. Thanks for the advices. Patches will be ready soon. I'm very glad to get involved into these cool stuffs.

src/cloud/provision/freedom-module.ts
jamesliu96: @trevj I tried but there comes an error message:
```
error TS2322: Type 'Promise<{}>' is not assignable to type 'Promise<void>'.
  Type '{}' is not assignable to type 'void'.
```
So I changed it to `Promise<{}>`

jamesliu96:
@trevj Ok, now it seems to be fixed :smiley: I'll be working with Ray with the cloud provision part next, with more public cloud service providers supported.

src/cloud/provision/freedom-module.ts
trevj: I've run into this too. For the final return (currently line 333) you need to explicitly cast to Promise<void>, i.e.:
```typescript
return Promise.resolve<void>();
```

Then it should work.

trevj:
:+1:, once the `Promise.resolve<void>();` is in.

Thanks again!

src/cloud/provision/freedom-module.ts
jamesliu96: @trevj Thanks! I should have read more other modules to just learn the pattern.

jamesliu96: Still
```
error TS2453: The type argument for type parameter 'U' cannot be inferred from the usage. Consider specifying the type arguments explicitly.
  Type argument candidate 'Object' is not a valid type argument because it is not a supertype of candidate 'void'.
```
so I determined to do the cast outside.

trevj: So, the return type should be `Promise<void>` and the return statement should be `return Promise.resolve<void>`. If you can't get that working it's fine, we can revisit later. Still :+1: .

src/cloud/provision/freedom-module.ts
```diff
@@ -0,0 +1,357 @@
...
+  /**
+   * Properly configure Digital Ocean with a single droplet of name:name
+   * Assumes we already have oAuth token and  SSH key in this.state_
+   * This method will use this.waitDigitalOceanActions_() to wait until all actions complete
+   * before resolving
+   * @param {String} name of droplet
+   * @return {Promise.<void>} resolves on success, rejects on failure
+   */
+  private setupDigitalOcean_ = (name: string) : Promise<void> => {
+    return new Promise<void>((F, R) => {
```
jamesliu96: @trevj so I did it here. will it be fine in this case?

jamesliu96: This passes all compilations and tests.

trevj: Perfect!

jamesliu96: and the doc was fixed also. :smiley:

trevj:
:+1:!

ryscheng:
Awesome! This is great :-)

Trevor, i think you have to merge, since James doesn't have direct commit access yet

trevj:
Ah! Thanks @ryscheng. Hadn't realised. Merging now.

PR #334, opened 2015-12-10
implement inviteUser in the cloud social provider

trevj:
@lucyhe, could really use your advice here:
-  is `inviteUser` invoked with the ID of the user to which you're offering access?
- right now this returns an invite code as it should be distributed amongst users; is that what uProxy expects or should I be returning an object and allowing uProxy base64 encode it (or something)?

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/334)

### DIFF
--- a/src/cloud/social/provider.ts
+++ b/src/cloud/social/provider.ts
@@ -22,6 +22,10 @@ const ZORK_PORT = 9000;
 // Key under which our contacts are saved in storage.
 const STORAGE_KEY = 'cloud-social-contacts';
 
+// Admin users can issue invites.
+const ADMIN_USERNAME = 'giver';
+const REGULAR_USERNAME = 'getter';
+
 // Credentials for accessing a cloud instance.
 // The serialised, base64 form is distributed amongst users.
 // TODO: add (private) keys, for key-based auth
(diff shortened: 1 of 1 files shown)


lucyhe:
Since the codes aren't related to any contacts in the roster, I dont think inviteUser should expect a userId. Right now we have a function [getInviteUrl()](https://github.com/uProxy/uproxy/blob/dev/src/generic_core/social.ts#L748) that calls inviteUser with an empty string, and returns a www.uproxy.org/invite/... url. I think this is probably what we want to use for cloud (at least for now)!

If possible, it would be best for inviteUser to return an object: { networkData :string }. uProxy will base64 this to turn it into the URL. Lets chat if that doesn't seem to work well with what you have so far; we can figure something else out!

trevj:
OK, as discussed:
 * uProxy will pass a `clientId` to inviteUser, as per-Github
 * `inviteUser` now returns an object, with a `networkData` field that uProxy will base64 encode

lucyhe:
:+1:

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-lib/334#-:-K5I-5D6IVxjBURAOqRc:676269983)*

PR #338, opened 2015-12-15
Accept 403 responses in the uproxy.org 404 integration test.

bemasc:
This started failing after we started using cloudflare for
uproxy.org

Fixes https://github.com/uProxy/uproxy/issues/2134

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/338)


trevj:
:+1: Thanks!

trevj:
ping!

Does this fix the tests? Let's submit, I will need to release later this week.

jab:
@bemasc You had written "a 301 would make the test fail" and then "I think it would make more sense to change the target server to something with more relevant behavior." If it's still worth it please chime in. In the meantime, happy to add logic to allow a 301 or 302 if that's helpful.

PR #339, opened 2015-12-17
Clean up the Gruntfile

jpevarnek:
Reopening #300

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/339)


trevj:
Thanks for updating this! It does seem like an improvement. Just a small couple of comments for now. Did you try building uProxy when it's linked against this?

jpevarnek:
I know I tested building a uproxy that relied on these rules and I'm almost
positive that it will work with old versions of uProxy.  Give me a minute
to double-check that though.

jpevarnek:
Confirmed: all changes are backwards compatible.

src/build-tools/common-grunt-rules.ts
```diff
@@ -215,4 +231,72 @@ export class Rule {
...
+   *   Lists all tests under the src/ directory
+   * getTests('src', 'generic_ui/scripts');
+   *   Lists all tests under the generic_ui/scripts directory, all paths
+   *   relative to src/
+   * getTests('src', undefined, ['integration-tests']);
+   *   Lists all the tests under src/ ignoring anything named integration-tests
+   */
+  public getTests(rootDir :string, current?:string, ignore?:string[]) {
+    var tests :string[] = [];
+    if (typeof current === 'undefined') {
```
trevj: It looks like `current` and `ignore` could have default values instead?

src/build-tools/common-grunt-rules.ts
```diff
@@ -215,4 +231,72 @@ export class Rule {
...
+   *
+   * getTests('src');
+   *   Lists all tests under the src/ directory
+   * getTests('src', 'generic_ui/scripts');
+   *   Lists all tests under the generic_ui/scripts directory, all paths
+   *   relative to src/
+   * getTests('src', undefined, ['integration-tests']);
+   *   Lists all the tests under src/ ignoring anything named integration-tests
+   */
+  public getTests(rootDir :string, current?:string, ignore?:string[]) {
```
trevj: I would be strongly inclined to drop the `ignore` parameter, in favour of having the caller perform any filtering. What do you think? I reckon it'll make both this function and the call to it, in the Gruntfile, simpler.

src/build-tools/common-grunt-rules.ts
```diff
@@ -95,6 +96,21 @@ export class Rule {
     }
   }
 
+  public jasmineSingleSpec(file :string) :JasmineRule {
```
trevj: How `jasmineSpec`?

jpevarnek: I didn't want to do a name conflict with the current rule since the behaviour changed

src/build-tools/common-grunt-rules.ts
jpevarnek: Oh, huh, I didn't realize that was possible in Typescript.  Thanks!

src/build-tools/common-grunt-rules.ts
trevj: Ah, I missed that :-)

OK, this name seems fine!

src/build-tools/common-grunt-rules.ts
jpevarnek: Can you tell me a bit more about how you think that will clean things up?

To explain my reasoning, it's three extra lines here and makes it very clear how those are handled.  Doing the same thing within the gruntfile would require filtering through the output array and removing all entries that have the string(s) in the filename.  While it might not be needed in every call, I think having it as an optional last parameter would be overall simpler.

trevj: As we discussed, let's leave that final parameter since you reckon it will be useful to many callers.

trevj:
:+1:

jpevarnek:
Merging, this will require at least a minor version bump in the next release.

PR #341, opened 2015-12-17
Ensure that the Churn candidates' MID matches the SDP.

bemasc:
Fixes https://github.com/uProxy/uproxy/issues/2143

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/341)


trevj:
Great. This class continues to spiral out of control but we can tackle that later. Some tests would be good  but I just tried this with my own Firefox-based cloud instance and it seems to fix the issue so I'm going to merge this and roll a uproxy-lib release.

PR #342, opened 2015-12-18
UserProfile should include status that indicates if a Cloud instance is owned by or shared with the local user.

lucyhe:
Depends on: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/7263

uProxy will check the UserStatus to determine if a user should see a cloud instance in their Share tab (and therefore see a URL for sharing that cloud instance with others).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/342)


src/cloud/social/provider.ts
```diff
@@ -112,10 +120,17 @@ function makeClientState(address: string): freedom.Social.ClientState {
 // To see how these fields are handled, see
 // generic_core/social.ts#handleUserProfile in the uProxy repo. We omit
 // the status field since remote-user.ts#update will use FRIEND as a default.
-function makeUserProfile(address: string): freedom.Social.UserProfile {
+function makeUserProfile(
+    address: string,
+    username :string): freedom.Social.UserProfile {
+  var status = UserStatus.CLOUD_INSTANCE_SHARED_WITH_LOCAL;
+  if (username == ADMIN_USERNAME) {
```
trevj: `===`

src/cloud/social/provider.ts
```diff
@@ -58,6 +58,14 @@ interface SavedContact {
   description?: string;
 }
 
+var UserStatus = {
```
trevj: It looks like you're extending src/interfaces/social.ts#UserStatus from the uProxy repo, is that right?

src/cloud/social/provider.ts
lucyhe: done

src/cloud/social/provider.ts
lucyhe: Yes! I have parallel changes in progress in a uproxy branch: https://github.com/uProxy/uproxy/tree/lucyhe-cloudinvite

trevj: Ah, great :-) In that case, how about:
 * add a comment explaining what this duplicates
 * making this an enum too
 * make the value names match

Separately, what do you think of moving the cloud social provider to the uproxy repo? Seems like this sort of thing will happen a lot.

lucyhe: Ah thanks for catching!! I worked off of some older code.

Hmm I hadn't considered moving this into uproxy since we've been keeping social providers separate until now. I agree that there should be a solution to copy/pasting code from uproxy into different social providers though (we have this code in github too).

dborkan: This user.status field is kinda annoying.  I entered https://github.com/uProxy/uproxy/issues/2158 to clean this up.  I think the slightly longer term solution is to convert user.status to a string field, and then emit string values like "FRIEND" - this is the same as client.status and the way FreedomJS expects things to work (a little annoying in TypeScript though because TypeScript enums usually evaluate to numbers... actually they are maps like ```{"FRIEND": 0, 0: "FRIEND"}```).

I don't have a strong feeling about moving cloud to the uProxy repo, but I'm not sure it's worth doing just for this enum.

trevj: Great, thanks for changing this.

Regarding moving this code to the uProxy repo, it seems like most of the changes we're making  here are in response to new UI and it's generally a pain to test changes cross-repo and release uproxy-lib just for cloud social provider changes (my dream: one giant repo). Anyway, it's not pressing so just food for thought for now.

trevj:
:+1:

PR #344, opened 2016-01-04
fix issue where inviteUser rejects

trevj:
Fixes an issue by which `inviteUser` rejects unless the cloud instance has been added since last login.

Basically, we should check `this.savedContacts_` for the presence of an invite with an ADMIN username and then `reconnect`. Oops, thanks for spotting!

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/344)


lucyhe:
This is almost working as expected in my testing!
The situation I'm encountering:
Upon login, inviteUser is called to get an inviteUrl, but still exits early because it cant find the clientId in savedContacts. However, if I start proxying and then refresh the UI, inviteUser will work. (sendMessage calls reconnect_, and the correct entry is added to savedContacts.)

trevj:
OK, I think `inviteUser` probably shouldn't be called at login -- let's chat later :-)

trevj:
I've just been testing this with your pull request and inviting seems to work now, even if I haven't added the cloud instance since logging into cloud.

lucyhe:
awesome, sounds good! :+1:

PR #345, opened 2016-01-04
Update the address of the NAT probe server

jpevarnek:
IP address is now an EC2 elastic IP so we should be able to move it even
if we have to shut down the server.

Fixes uproxy/uproxy#2157

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/345)


dborkan:
:+1:

trevj:
Yay!

:+1:

jpevarnek:
I just changed the IP address to one under the shared uproxy AWS instead of mine, no other changes here.

PR #347, opened 2016-01-12
add digital ocean regions parameter, some indenting

agallant:
Fix https://github.com/uProxy/uproxy/issues/2049

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/347)

### DIFF
--- a/src/cloud/digitalocean/provisioner.ts
+++ b/src/cloud/digitalocean/provisioner.ts
@@ -62,9 +62,6 @@ class Provisioner {
       // Get SSH keys
     }).then((keys: KeyPair) => {
       this.state_.ssh = keys;
-      if (typeof region === "undefined") {
-        region = "nyc3";
-      }
       return this.setupDigitalOcean_(name, region);
       // Setup Digital Ocean (SSH key + droplet)
     }).then(() => {
--- a/src/cloud/digitalocean/provisioner.ts
+++ b/src/cloud/digitalocean/provisioner.ts
@@ -7,6 +7,10 @@ declare var forge: any;
 
 const POLL_TIMEOUT: number = 5000; //milliseconds
 
+// To see available regions:
+// https://developers.digitalocean.com/documentation/v2/#list-all-regions
+const DEFAULT_REGION: string = "nyc3";
+
 const STATUS_CODES: { [k: string]: string; } = {
   "START": "Starting provisioner",
   "OAUTH_INIT": "Initializing oauth flow",
@@ -48,7 +52,8 @@ class Provisioner {
    * @param {String} region to create VM in
    * @return {Promise.<Object>}
    */
-  public start = (name: string, region?: string) : Promise<Object> => {
+  public start = (name: string, region: string = DEFAULT_REGION) :
+      Promise<Object> => {
     this.sendStatus_("START");
     // Do oAuth
     return this.doOAuth_().then((oauthObj: any) => {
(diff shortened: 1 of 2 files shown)


soycode:
One thought I'm considering - it may be worth making the `region` parameter an object to ensure that it is useful across cloud providers. A string parameter may be good enough most of the time, but more complex providers may need multiple parameters specified.

src/cloud/digitalocean/provisioner.ts
```diff
@@ -45,22 +45,26 @@ class Provisioner {
    * One-click setup of a VM
    * See freedom-module.json for return and error types
    * @param {String} name of VM to create
+   * @param {String} region to create VM in
```
trevj: Can you add a link to the canonical list of region names?

src/cloud/digitalocean/provisioner.ts
```diff
@@ -45,22 +45,26 @@ class Provisioner {
...
       this.state_.oauth = oauthObj;
       return this.getSshKey_(name);
-    // Get SSH keys
+      // Get SSH keys
     }).then((keys: KeyPair) => {
       this.state_.ssh = keys;
-      return this.setupDigitalOcean_(name);
-    // Setup Digital Ocean (SSH key + droplet)
+      if (typeof region === "undefined") {
+        region = "nyc3";
```
trevj: Function parameters can have defaults, i.e.
```typescript
public start = (name: string, region: string = 'nyc3'): Promise<Object> => {
  // do stuff
}
```
You could also extract `nyc3` as a `const`.

src/cloud/digitalocean/provisioner.ts
```diff
@@ -149,18 +153,18 @@ class Provisioner {
       this.sendStatus_("OAUTH_INIT");
       oauth.initiateOAuth(REDIRECT_URIS).then((obj: any) => {
         var url = "https://cloud.digitalocean.com/v1/oauth/authorize?" +
-                  "client_id=c16837b5448cd6cf2582d2c2f767cfb7d11844ec395a91b43f26ca72513416c8&" +
-                  "response_type=token&" +
-                  "redirect_uri=" + encodeURIComponent(obj.redirect) + "&" +
-                  "state=" + encodeURIComponent(obj.state) + "&" +
-                  "scope=read%20write";
+          "client_id=c16837b5448cd6cf2582d2c2f767cfb7d11844ec395a91b43f26ca72513416c8&" +
```
trevj: 4 spaces indentation

src/cloud/digitalocean/provisioner.ts
```diff
@@ -149,18 +153,18 @@ class Provisioner {
...
+          "state=" + encodeURIComponent(obj.state) + "&" +
+          "scope=read%20write";
         return oauth.launchAuthFlow(url, obj);
       }).then((responseUrl: string) => {
         var query = responseUrl.substr(responseUrl.indexOf('#') + 1),
-            param: string,
-            params: { [k: string]: string } = {},
-            keys = query.split('&'),
-            i = 0;
+        param: string,
```
trevj: 4 spaces indentation (we don't use this capability of `var` anywhere else in uproxy-lib but I think 4 is more appropriate)

src/cloud/digitalocean/provisioner.ts
```diff
@@ -220,7 +224,7 @@ class Provisioner {
    * @return {Promise.<Object>} - JSON object of response body
    */
   private doRequest_ = (method: string, actionPath: string, body?: string) :
-      Promise<Object> => {
+  Promise<Object> => {
```
trevj: 4 spaces indentation

trevj:
Thanks. :+1:

I think we almost certainly will want to specify the config with an object - that parameter list is bound to grow quite a bit - but I think sticking with parameters is fine for now, we can draft a more general interface for these deployment modules once we have >1.

soycode:
Addressed comments I think, let me know if you want any other tweaks. And agreed re: the region parameterization, a string should be "good enough" for awhile at least.

trevj:
I think you forgot two comments?:
 * add a default value for the region param
 * link to the list of region names

soycode:
I set the `DEFAULT_REGION` as a constant towards the top, and put the comment there - I can move it to the function comment block if you think that's clearer.

trevj:
Sorry, missed the comment at the top! Great. But you don't need the check now at line 65 -- the variable will always be initialised.

soycode:
Missed that, good catch.

trevj:
:+1:

PR #349, opened 2016-01-21
add CloudInstaller, to install uProxy on a server via SSH

trevj:
This adds a class which will install uProxy on a server via SSH. The process is as similar as possible to a manual install (being described in an upcoming blog post). I've also extended the deployer sample app to use it.

Note that there is some weirdness with brand new Digital Ocean instances: for several minutes after creation, network access is flaky (`git`, `curl`, `host`, etc., fail or timeout or run extremely slowly). Still trying to figure that out.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/349)

### DIFF
--- a/src/cloud/install/freedom-module.json
+++ b/src/cloud/install/freedom-module.json
@@ -1,0 +1,26 @@
+{
+  "name": "Cloud Installer",
+  "description": "Installs uProxy in the cloud, via SSH.",
+  "app": {
+    "script": [
+      "freedom-module.static.js"
+    ]
+  },
+  "provides": [ "cloudinstall" ],
+  "default": "cloudinstall",
+  "api": {
+    "cloudinstall": {
+      "install": {
+        "type": "method",
+        "value": ["string", "number", "string", "string"],
+        "ret": "string"
+      },
+      "err": {
+        "message": "string"
+      }
+    }
+  },
+  "permissions": [
+    "core.tcpsocket"
+  ]
+}
--- a/src/cloud/install/installer.ts
+++ b/src/cloud/install/installer.ts
@@ -54,8 +54,6 @@ class CloudInstaller {
             return;
           }
           stream.on('end', () => {
-            log.debug('exec stream closed');
-            // TODO: is it correct to reject if we haven't already resolved?
             R({
               message: 'invitation URL not found'
             });
@@ -73,14 +71,13 @@ class CloudInstaller {
         // This occurs when:
         //  - user supplies the wrong username or password
         //  - host cannot be reached, e.g. non-existant hostname
-        log.warn('failed to connect: %1', e);
         R({
           message: 'could not login: ' + e.message
         });
       }).on('end', () => {
         log.debug('end');
       }).on('close', (hadError: boolean) => {
-        log.debug('close');
+        log.debug('close (hadError: %1)', hadError);
       }).connect(connectConfig);
     });
   }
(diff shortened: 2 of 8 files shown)


src/cloud/install/installer.ts
```diff
@@ -0,0 +1,85 @@
...
+        });
+      }).on('error', (e: Error) => {
+        // This occurs when:
+        //  - user supplies the wrong username or password
+        //  - host cannot be reached, e.g. non-existant hostname
+        R({
+          message: 'could not login: ' + e.message
+        });
+      }).on('end', () => {
+        log.debug('end');
```
soycode: Do this (and 'close') below need to actually do any cleanup?

src/cloud/install/freedom-module.json
```diff
@@ -0,0 +1,26 @@
...
+        "value": ["string", "number", "string", "string"],
+        "ret": "string"
+      },
+      "err": {
+        "message": "string"
+      }
+    }
+  },
+  "permissions": [
+    "core.tcpsocket"
```
soycode: Is the ssh2 package successfully depending on this? I don't see any explicitly instantiated socket, though clearly the dependency will be needed.

src/cloud/install/installer.ts
```diff
@@ -0,0 +1,85 @@
...
+            return;
+          }
+          stream.on('end', () => {
+            R({
+              message: 'invitation URL not found'
+            });
+          }).on('data', function(data: Buffer) {
+            const output = data.toString();
+            log.debug('STDOUT: %1', output);
+            if (output.indexOf(INVITATION_URL_PREFIX) === 0) {
```
soycode: You can possibly use `output.startsWith(INVITATION_URL_PREFIX)` instead (according to http://kangax.github.io/compat-table/es6/#String.prototype.startsWith it's supported in reasonably new FF, Chrome, Edge, and Node - but not IE11).

soycode:
Few nits, and re: new instance weirdness I have similar experience on Linode (you have to wait a few minutes after first boot for the StackScript to finish installing docker, and it's not exactly clear how long to wait). This probably isn't universally solvable, but coming up with a good heuristic (backed by reasonable error surfacing, or maybe just auto-retry a few times) is probably the best we can do.

src/cloud/install/installer.ts
trevj: Yes. Great catch. I examined the ssh2 docs and ran some tests here locally and I believe this now correctly closes the SSH connection in all cases. Thanks!

src/cloud/install/freedom-module.json
trevj: The node -> freedom shims at `src/cloud/social/shim` depend on it.

src/cloud/install/installer.ts
trevj: Agreed, but I deliberately did this in order to survive small changes to the install script, e.g. `cloud invite: xxx` changing to `URL: xxx`. I've added a comment to explain this.

trevj:
Great to know that's the case on Linode, too. We can probably have the DigitalOcean ping the server or something until we get a response...I filed an issue to investigate further:
https://github.com/uProxy/uproxy/issues/2190

src/cloud/install/freedom-module.json
soycode: Gotcha, thanks.

soycode:
:+1:

PR #350, opened 2016-01-22
polish for cloud social provider SSH use

trevj:
Applying some polish, based on what I learned while working on the installer:
https://github.com/uProxy/uproxy-lib/pull/349

Main two things:
 * reject with `{message: 'xxx'}` objects (for freedom)
 * remove some handlers on the SSH objects which don't actually get called in practise

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/350)

### DIFF
--- a/src/cloud/social/provider.ts
+++ b/src/cloud/social/provider.ts
@@ -538,7 +538,7 @@ class Connection {
         log.debug('%1: connection end', this.name_);
         this.setState_(ConnectionState.TERMINATED);
       }).on('close', (hadError: boolean) => {
-        log.debug('%1: connection close, with%1 error', this.name_, (hadError ? '' : 'out'));
+        log.debug('%1: connection close, with%2 error', this.name_, (hadError ? '' : 'out'));
         this.setState_(ConnectionState.TERMINATED);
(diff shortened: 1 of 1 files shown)


trevj:
BTW, I hope the promise rejection fixes may help with this issue:
https://github.com/uProxy/uproxy/issues/2183

src/cloud/social/provider.ts
```diff
@@ -26,6 +26,9 @@ const STORAGE_KEY = 'cloud-social-contacts';
 const ADMIN_USERNAME = 'giver';
 const REGULAR_USERNAME = 'getter';
 
+// Timeout for establishing an SSH connection.
+const CONNECT_TIMEOUT_MS = 10000;
```
soycode: It looks like the default is 20000, just curious the reason for shortening it.

trevj: 20s just seemed like a long time to be sitting there waiting for an error.

soycode: :+1: fair enough.

soycode:
:+1:

PR #354, opened 2016-02-04
some tidy for ssh key generation in digitalocean module

trevj:
Been playing with this from the frontend...some code tidyup for the SSH key generation part.

Much more to come.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/354)

### DIFF
--- a/src/cloud/digitalocean/provisioner.ts
+++ b/src/cloud/digitalocean/provisioner.ts
@@ -16,8 +16,10 @@ const STATUS_CODES: { [k: string]: string; } = {
   "OAUTH_INIT": "Initializing oauth flow",
   "OAUTH_ERROR": "Error getting oauth token",
   "OAUTH_COMPLETE": "Got oauth token",
-  "SSHKEY_RETRIEVED": "Retrieved SSH keys from storage",
+  "SSHKEY_RETRIEVED": "Loaded SSH keys from storage",
+  "GENERATING_SSHKEY": "Generating new SSH keys",
   "SSHKEY_GENERATED": "Generated new SSH keys",
+  "SSHKEY_SAVED": "Saved SSH keys to storage",
   "CLOUD_FAILED": "Failed to complete cloud operation",
   "CLOUD_INIT_ADDKEY": "Starting to add SSH key to cloud account",
   "CLOUD_DONE_ADDKEY": "Done adding SSH key to cloud account",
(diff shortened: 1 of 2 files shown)


src/cloud/digitalocean/provisioner.ts
```diff
@@ -182,38 +191,52 @@ class Provisioner {
...
+      }
+
+      this.sendStatus_('GENERATING_SSHKEY');
+      try {
+        const result = Provisioner.generateKeyPair_();
+        this.sendStatus_('SSHKEY_GENERATED');
+        return Promise.all([
+          storage.set(publicKeyIndex, result.public),
+          storage.set(privateKeyIndex, result.private)
+        ]).then((ignored: any) => {
```
soycode: I think you could just omit the argument, i.e. `then(() => {`.

src/cloud/digitalocean/provisioner.ts
```diff
@@ -120,11 +122,18 @@ class Provisioner {
...
    */
   private static generateKeyPair_ = () : KeyPair => {
-    "use strict";
-    var pair = forge.pki.rsa.generateKeyPair({bits: 2048, e: 0x10001});
-    var publicKey = forge.ssh.publicKeyToOpenSSH(pair.publicKey, '');
-    var privateKey = forge.ssh.privateKeyToOpenSSH(pair.privateKey, '');
-    return { public: publicKey, private: privateKey };
+    const pair = forge.pki.rsa.generateKeyPair({
+      bits: 2048,
+      e: 0x10001
```
soycode: Looks like this is the default parameter, so it's worth considering omitting.

trevj: Yes. Done.

src/cloud/digitalocean/provisioner.ts
trevj: I can, and it will compile, but I'd rather make it explicit that while we know there's a returned object, we don't need it. Plus, TypeScript has been getting better and better at recognising things like extra fields in objects -- this may start raising an error in future versions.

trevj:
Thanks for the review!

src/cloud/digitalocean/provisioner.ts
soycode: SGTM.

soycode:
:+1:

PR #358, opened 2016-03-03
support v2 invite output from issue_invite.sh

trevj:
Supports the v2 format introduced introduced here:
https://github.com/uProxy/uproxy-docker/pull/55

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/358)

### DIFF
--- a/src/cloud/social/provider.ts
+++ b/src/cloud/social/provider.ts
@@ -564,10 +564,31 @@ class Connection {
     return this.exec_('cat /banner');
   }
 
-  // Returns a base64-decoded, deserialised invite code.
+  // Returns an object to be used as the networkData field in an invitation URL.
   public issueInvite = (): Promise<Object> => {
-    return this.exec_('sudo /issue_invite.sh').then((inviteCode: string) => {
-      return JSON.parse(new Buffer(inviteCode, 'base64').toString());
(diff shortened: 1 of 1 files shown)


src/cloud/social/provider.ts
```diff
@@ -564,10 +564,31 @@ class Connection {
...
+      //   {
+      //     "networkName": "Cloud"
+      //     "networkData": "<the above JSON blob>"
+      //   }
+      //
+      // Yes, that was weird. To support those older servers we first try to
+      // base64-decode the output. If we discover base64, we return the decoded
+      // object's "networkData" field - otherwise, return the output verbatim.
+      try {
+        const decodedOutput = JSON.parse(new Buffer(output, 'base64').toString());
```
soycode: Are we sure this will throw an error for the new format? An alternative way to do this instead of try/catch may be to see if the "v=2" parameter is present.

trevj:
Closing this for now, since I don't think uProxy will be issuing additional invites for users at launch.

dborkan:
I'm actually making changes here so that issue_invite.sh just outputs the 3 fields needed for uProxy to generate the invite (host, user, key).  Once this is done, we will have uProxy doing v2 invites and issue_invite.sh won't need to worry about the invite format anymore

PR #359, opened 2016-03-08
make default image and size match the blog post

trevj:
Testing now for sure with your https://github.com/uProxy/uproxy/pull/2294 but I think this makes it work end-to-end :-)

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/359)


dborkan:
:+1:

PR #360, opened 2016-03-09
fix for newline handling on SSH input

trevj:
Oops, we always did this for telnet and I never thought to do it for SSH. Fixes the issue in which invites (crazy long) were being split.

---

This change is [Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/360)


dborkan:
:+1:

PR #361, opened 2016-03-09
Cleanup cloud invite code to not require SSH access

dborkan:
Cleanup cloud invite code to not require SSH access.  Tested by accepting and inviting cloud instances in Chrome

---

This change is [Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/361)

### DIFF
--- a/src/cloud/social/provider.ts
+++ b/src/cloud/social/provider.ts
@@ -43,6 +43,8 @@ interface Invite {
   pass?: string;
   // Private key, base64-encoded.
   key?: string;
+  // Is Admin.
+  isAdmin?: boolean;
 }
 
 // Type of the object placed, in serialised form, in storage
@@ -130,11 +132,9 @@ function makeClientState(address: string): freedom.Social.ClientState {
 // the status field since remote-user.ts#update will use FRIEND as a default.
 function makeUserProfile(
     address: string,
-    username :string): freedom.Social.UserProfile {
-  var status = UserStatus.CLOUD_INSTANCE_SHARED_WITH_LOCAL;
-  if (username === ADMIN_USERNAME) {
-    status = UserStatus.CLOUD_INSTANCE_CREATED_BY_LOCAL;
-  }
+    isAdmin :boolean): freedom.Social.UserProfile {
+  var status = isAdmin ? UserStatus.CLOUD_INSTANCE_CREATED_BY_LOCAL :
+      UserStatus.CLOUD_INSTANCE_SHARED_WITH_LOCAL;
   return {
     userId: address,
     name: address,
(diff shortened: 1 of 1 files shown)


src/cloud/social/provider.ts
```diff
@@ -36,13 +32,13 @@ interface Invite {
   // Hostname or IP of the cloud instance.
```
trevj: Guess we can remove the TODO, too :-)

src/cloud/social/provider.ts
```diff
@@ -36,13 +32,13 @@ interface Invite {
...
+  host: string;
   // Username.
-  user?: string;
-  // Password.
-  pass?: string;
+  user: string;
   // Private key, base64-encoded.
-  key?: string;
+  key: string;
+  // Is Admin.
```
trevj: How about:
```js
// True iff uProxy has root access on the server, i.e. uProxy deployed it.
```

src/cloud/social/provider.ts
```diff
@@ -363,26 +356,17 @@ export class CloudSocialProvider {
...
-        message: 'unknown cloud instance ' + clientId
-      });
-    }
-    if (this.savedContacts_[clientId].invite.user !== ADMIN_USERNAME) {
-      return Promise.reject({
-        message: 'user is logged in as non-admin user ' +
-            this.savedContacts_[clientId].invite.user
-      });
+    if (!(host in this.savedContacts_)) {
+      return Promise.reject({message: 'unknown cloud instance ' + host});
```
trevj: Keep message on its own line.

src/cloud/social/provider.ts
```diff
@@ -363,26 +356,17 @@ export class CloudSocialProvider {
   // social2
   ////
 
-  // Returns a new invite code for the specified server.
-  // Rejects if the user is not logged in as an admin or there
-  // is any error executing the command.
-  // TODO: typings for invite codes
-  public inviteUser = (clientId: string): Promise<Object> => {
+  // Returns the invite code for the specified server.
+  public inviteUser = (host: string): Promise<Object> => {
```
trevj: We can make the return type `Invite`, right?

trevj:
Thanks for the cleanup! Just a couple of style things.

src/cloud/social/provider.ts
dborkan: I tried that, TypeScript compilation gave an error because Invite is private to this file (not exported)

trevj: OK. There's probably no advantage to exporting it, right, this should just be an opaque object to the rest of the system.

Since it's intended to be re-consumed by other methods in this class I suggest casting it when you return, i.e.
```typescript
return Promise.resolve(<Invite>{
  host: invite.host,
  user: invite.user,
  key: invite.key
});
```

dborkan:
Just made those changes, PTAL

trevj:
:+1:!

PR #362, opened 2016-03-10
Update installer to return invite object instead of URL

dborkan:
This change is [Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/362)


trevj:
:+1:

PR #363, opened 2016-03-15
fix issues with newlines on SSH connections

trevj:
Fixes the issue in which cloud contacts can't be added or updated, due to newline issues with `cat /banner`. Also for Zork,where this may have been an issue.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/363)


dborkan:
:+1:

PR #365, opened 2016-03-16
Added OAuth to provisioner stop method

gitlaura:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/365)


src/cloud/digitalocean/provisioner.ts
```diff
@@ -104,8 +105,13 @@ class Provisioner {
    * @param {String} name of VM to create
    * @return {Promise.<Object>}
```
trevj: Let's quickly describe the format of the return type while we're at it...is it just:
```js
{
  droplet: <droplet id, as a string>
}
```
?

src/cloud/digitalocean/provisioner.ts
```diff
@@ -104,8 +105,13 @@ class Provisioner {
...
    */
-  public stop = (name: string) : Promise<Object> => {
-    return this.doRequest_('GET', 'droplets').then((resp: any) => {
+  public stop = (name: string): Promise<Object> => {
+    this.sendStatus_('STOP');
+    return this.doOAuth_().then((oauthObj: any) => {
```
trevj: I believe `Object` is preferable to `any`, unless there's a special reason to return `any` (and below).

trevj:
Sorry for the delay!

src/cloud/digitalocean/provisioner.ts
gitlaura: Updated the return type to Promise<void> since there isn't anything to return.

src/cloud/digitalocean/provisioner.ts
gitlaura: It appears that using Object does not compile because you can't call `.droplets` on type Object. I assume this is why `resp :any` has been used throughout provisioner.ts. Any suggestions?

gitlaura:
I updated provisioner.ts to separate the `stop` and `destroyServer_` methods so that we can call `destroyServer_` if there's a failure when creating a cloud server.

src/cloud/digitalocean/provisioner.ts
```diff
@@ -99,26 +101,48 @@ class Provisioner {
...
+
+  /**
+   * Destroys cloud server; assumes OAuth has already been completed
+   * @todo DELETE request does not return a response body so we
+   * need to check that the response header indicates success (204)
+   * @param {String} droplet name, as a string
+   * @return {Promise.<void>}
+   */
+  private destroyServer_ = (name: string): Promise<void> => {
+    if (this.state_.oauth) {
```
trevj: Rather than wrapping the whole function in a try/catch block you could make this a precondition-like block:
```typescript
if (!this.state_.oauth) {
  return Promise.reject({
    'errcode': 'OAUTH_ERR',
    'message': 'Cannot destroy server - not logged into digital ocean'
  });
  ...
  ...
}
```

Actually...since this is a private method only called by `stop`, do you really need this check?

trevj:
BTW, am investigating the Travis failures now...

src/cloud/digitalocean/provisioner.ts
gitlaura: Agree that we don't need it. Made this change.

trevj:
:+1:

PR #371, opened 2016-03-28
retry for SSH connections, with a re-useable library

trevj:
This fixes the case where, immediately after an install, the SSH container isn't **quite** ready to accept connections.

Also some refactoring, solving https://github.com/uProxy/uproxy/issues/2295.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/371)


dborkan:
:+1:

PR #373, opened 2016-03-29
Checking for 204 response status when deleting cloud server

gitlaura:
doRequest returns the response body for a request, but DigitalOcean doesn't return a response body when deleting a server. Instead we need to check for a 204 response status.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/373)


trevj:
:+1:

Good catch!

PR #374, opened 2016-03-29
go back to the docker slug

trevj:
DigitalOcean's Docker slug has been updated, meaning we can now safely use it and avoid the time taken to install Docker on the droplet. Just tested this and was able to create a cloud server in seven minutes.

No changes required to the Docker scripts; I can update the blog post when this is in.

---
This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/374)


trevj:
@dborkan, FYI. Shaves many seconds off the install.

dborkan:
Awesome!

soycode:
:+1: - as an aside, Linode will still need to install docker when we get to it, but install_cloud.sh handles that as needed.

trevj:
Right! We can keep the Docker install check but this will speed the deploy process for most users. Thanks!

PR #376, opened 2016-03-31
fixes for SSH on firefox

trevj:
Two main differences:
1. Unlike on Chrome, connection failures are reported **instantly** if the SSH server is not ready. This means a simple retry loop fails in a matter of a few ms - so exponential backoff is required.
2. When this happens, no error is reported...weird. This has the effect of leaving the Promise unresolved forever; fortunately, the `end` and `close` events are still received. So, call the reject function when those events are received (this works because calling the reject function has no effect if the fulfill function has already been called).

With these two changes, I can successfully create a cloud server in Firefox.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/376)

### DIFF
--- a/src/cloud/install/installer.ts
+++ b/src/cloud/install/installer.ts
@@ -6,6 +6,7 @@
 import arraybuffers = require('../../arraybuffers/arraybuffers');
 import linefeeder = require('../../net/linefeeder');
 import logging = require('../../logging/logging');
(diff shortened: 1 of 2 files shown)


gitlaura:
Yay! Works for me on Firefox :+1:

PR #378, opened 2016-03-31
Allow inbound traffic from browser ports bound to the "ANY" interface.

bemasc:
Fixes https://github.com/uProxy/uproxy/issues/2334

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/378)


trevj:
Excellent! :+1:

bemasc:
Thanks!  Since I could never reproduce the issue, we might want to check it on your machine before committing.

trevj:
Good idea! I'll do this when I'm back tomorrow.

trevj:
It works! Yay! :+1:

PR #379, opened 2016-04-05
Checking for existing uProxy cloud server

gitlaura:
- Refactored common code into `getDroplet_()`
- Checking for existing uproxy cloud server in `start()` and returning unique error if server already exists
- Using `this.waitDigitalOceanActions_()` in `destroyServer()` to determine when the DELETE action has been completed (so that we can then create a new server if necessary)

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/379)

### DIFF
--- a/src/cloud/digitalocean/provisioner.ts
+++ b/src/cloud/digitalocean/provisioner.ts
@@ -71,9 +71,13 @@ class Provisioner {
     }).then((keys: KeyPair) => {
        // Get SSH keys
       this.state_.ssh = keys;
-      return this.getDroplet_(name);
+      return this.getDroplet_(name).catch((e: Error) => {
+        // Droplet does not exist so continue creating new server
+        return Promise.resolve<void>();
+      });
     }).then((resp: any) => {
-      if (resp.droplet) {
+      if (resp) {
+        // Droplet exists so return error
         return Promise.reject({
           'errcode': 'VM_AE',
           'message': 'Droplet ' + name + ' already exists'
@@ -132,12 +136,6 @@ class Provisioner {
       // Find and delete the server with the same name
       return this.getDroplet_(name);
     }).then((resp :any) => {
-      if (!resp.droplet) {
-        return Promise.reject({
-          'errcode': 'VM_DNE',
-          'message': 'Droplet ' + name + ' doesnt exist'
-        });
-      }
       this.state_.cloud = {};
       this.state_.cloud.vm = resp.droplet;
       return this.doRequest_('DELETE', 'droplets/' + resp.droplet.id);
@@ -170,8 +168,9 @@ class Provisioner {
           });
         }
       }
-      return Promise.resolve({
-        droplet: undefined
+      return Promise.reject({
+        'errcode': 'VM_DNE',
+        'message': 'Droplet ' + name + ' doesnt exist'
       });
     });
   }
--- a/src/cloud/digitalocean/provisioner.ts
+++ b/src/cloud/digitalocean/provisioner.ts
@@ -71,23 +71,20 @@ class Provisioner {
     }).then((keys: KeyPair) => {
        // Get SSH keys
       this.state_.ssh = keys;
-      return this.getDroplet_(name).catch((e: Error) => {
-        // Droplet does not exist so continue creating new server
-        return Promise.resolve<void>();
-      });
-    }).then((resp: any) => {
-      if (resp) {
-        // Droplet exists so return error
+      return this.getDroplet_(name).then((unused :any) => {
+        // Droplet exists so raise error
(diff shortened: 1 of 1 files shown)


src/cloud/digitalocean/provisioner.ts
```diff
@@ -127,17 +170,9 @@ class Provisioner {
           });
         }
       }
-      return Promise.reject({
-        'errcode': 'VM_DNE',
-        'message': 'Droplet ' + name + ' doesnt exist'
+      return Promise.resolve({
```
trevj: I think it's probably better to reject if the droplet doesn't exist; this will help simplify `destroyServer_` and `start`. What do you think?

trevj:
Nice! One comment.

src/cloud/digitalocean/provisioner.ts
gitlaura: Okay, I updated `getDroplet_`. Do you think this is better? Problem is that if `getDroplet_` fails then `start` should continue but if it returns a droplet then `start` should throw an error. So `start` wasn't really simplified, but I do think it's better if `getDroplet_` throws an error.

src/cloud/digitalocean/provisioner.ts
```diff
@@ -114,12 +127,40 @@ class Provisioner {
...
+    }).then(() => {
+      return Promise.resolve<void>();
+    });
+  }
+
+  /**
+   * Finds a droplet with this name
+   * @param {String} droplet name, as a string
+   * @return {Promise.<Object>}, returns droplet if exists
+   * or undefined if it does not exist
```
trevj: update comment :-)

src/cloud/digitalocean/provisioner.ts
```diff
@@ -114,12 +127,40 @@ class Provisioner {
...
+    }).then((resp: any) => {
+      if (resp.status.startsWith('204')) {
+        // Wait until server is deleted
+        this.sendStatus_('CLOUD_WAITING_VM');
+        return this.waitDigitalOceanActions_();
+      } else {
+        return Promise.reject(new Error('error deleting droplet'));
+      }
+    }).then(() => {
+      return Promise.resolve<void>();
```
trevj: I don't think you need this?

src/cloud/digitalocean/provisioner.ts
```diff
@@ -67,9 +68,21 @@ class Provisioner {
...
-      // Get SSH keys
     }).then((keys: KeyPair) => {
+       // Get SSH keys
       this.state_.ssh = keys;
+      return this.getDroplet_(name).catch((e: Error) => {
+        // Droplet does not exist so continue creating new server
+        return Promise.resolve<void>();
+      });
+    }).then((resp: any) => {
+      if (resp) {
```
trevj: I think `resp` will always be set, if the promise fulfills?

src/cloud/digitalocean/provisioner.ts
```diff
@@ -67,9 +68,21 @@ class Provisioner {
     return this.doOAuth_().then((oauthObj: any) => {
       this.state_.oauth = oauthObj;
       return this.getSshKey_(name);
-      // Get SSH keys
     }).then((keys: KeyPair) => {
+       // Get SSH keys
       this.state_.ssh = keys;
+      return this.getDroplet_(name).catch((e: Error) => {
```
trevj: Why don't you do something like:
```typescript
return this.getDroplet_(name).then((unused:any) => {
  // Droplet exists so raise an error.
  return Promise.reject({
    'errcode': 'VM_AE',
    'message': 'Droplet ' + name + ' already exists'
  });
}, (e:Error) => {
  return this.setupDigitalOcean_(name, region, image, size);
});
```
In fact, I don't see how it works right now.
 +        });

trevj:
Better I think! A couple of suggestions then we'll be good to go.

src/cloud/digitalocean/provisioner.ts
gitlaura: I like this better. Made this change.

src/cloud/digitalocean/provisioner.ts
gitlaura: It seemed to be working this way, but updated it based on your next comment.

src/cloud/digitalocean/provisioner.ts
gitlaura: Removed this.

src/cloud/digitalocean/provisioner.ts
gitlaura: Done!

trevj:
:+1: !

PR #381, opened 2016-04-07
move to grunt-browserify, with fix for node tcp shim

trevj:
Little code health exercise: move to grunt-browserify 5.x.

This revision is based on browserify 13 - the only incompatibility I've seen is that `toArrayBuffer` was removed; however, since Buffer now implements the Uint8Array interface I was able to just access its buffer field which is probably a little faster:
https://github.com/substack/node-browserify/blob/master/changelog.markdown#1300

BTW, some background on why they removed that method:
https://github.com/nodejs/node-v0.x-archive/issues/7609#issuecomment-42903457

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/381)

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "freedom-pgp-e2e": "^0.6.1",
     "freedom-xhr": "0.0.5",
     "grunt": "^0.4.5",
-    "grunt-browserify": "^4.0.0",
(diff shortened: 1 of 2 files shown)


jpevarnek:
:thumbsup:

PR #382, opened 2016-04-07
migrate from tsd to typings

trevj:
typings is the replacement for tsd:
https://github.com/typings/typings

My conclusion, having done this, is that typings is marginally less annoying than tsd. At least the config file no longer references Github commits.

Anyway, it's pretty much as before except I've opted for the "top-level" `.d.ts` rather than the individual `.d.ts`' as we've done to date, e.g.:

``` typescript
<reference path='../../../third_party/typings/browser.d.ts' />
```

I did this because it saves several lines and because I see mentions on typings' website of the directory layout within `typings/` changing between releases.

freedomjs was the only typing that presented a challenge: the type of `freedom` differs depending on whether your code runs inside or outside of a freedomjs module. The `freedom-core-env` and `freedom-module-env` typings exist for this purpose (they're essentially one-liners and you should include one or the other, depending on the environment) - but if you want the convenience of using the top-level import then you're out of luck.

My solution was to specify **only** `freedom` in `typings.json` and have each `freedom`-using source file specify the type of `freedom` immediately after imports, e.g.:

``` typescript
declare var freedom: freedom.FreedomInModuleEnv;
```

Horrible? Maybe...I'm loathe to lose the convenience of that top-level import just for freedomjs' weirdness, though. `egrep -ir 'declare (const|var) freedom' src/|wc -l` tells me that 47/117 source files needed the `declare` - and I think that will fall, in time. Would appreciate your thoughts on this.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/382)


jpevarnek:
This is fantastic, thanks for looking at it!  Quick note that we talked in person about browser.d.ts, but :thumbsup: at your discretion

---

Reviewed 93 of 93 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/arraybuffers/arraybuffers.ts, line 1 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/382#-KEqyS5aP6vMKdd3tpdL:-KEqyS5aP6vMKdd3tpdM:-563945045) ([raw file](https://github.com/uproxy/uproxy-lib/blob/27b1e1b2661b7698a9fadf008f01987184d74d81/src/arraybuffers/arraybuffers.ts#L1)):*
:thumbsup: :fireworks: 

---

*[src/integration-tests/socks-echo/base-spec.core-env.ts, line 34 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/382#-KEr-fzpfGJzQ4_482WP:-KEr-fzqZlf6UBFUcRNX:1174830885) ([raw file](https://github.com/uproxy/uproxy-lib/blob/27b1e1b2661b7698a9fadf008f01987184d74d81/src/integration-tests/socks-echo/base-spec.core-env.ts#L34)):*
Should we have a space before `any`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/uproxy-lib/382#-:-KEr0qFg-2TQXFi3M4y7:-2057633552)*

trevj:
Awesome, thanks for reviewing. I read through the docs again and it (still) seems like `browser.d.ts` is what we should be using.

Since this is a pretty majorly break change, I'll probably hold off on submitting until I get uproxy working well with it.

---

Review status: 90 of 93 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/integration-tests/socks-echo/base-spec.core-env.ts, line 34 \[r1\]](https://reviewable.io:443/reviews/uproxy/uproxy-lib/382#-KEr-fzpfGJzQ4_482WP:-KErLE25reZf3NiU9kc7:-330141709) ([raw file](https://github.com/uproxy/uproxy-lib/blob/884a668b281f9d343c9b37bd0489d238faa02999/src/integration-tests/socks-echo/base-spec.core-env.ts#L34)):*
Yes. Done, and in `slow.core-enc.spec.ts`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/uproxy-lib/382#-:-KErLK-OW5Pwy-BeDjGC:1537991750)*

jpevarnek:
Sounds good, let me know how I can help.

---

Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/uproxy-lib/382#-:-KF5JJEjUa6C09_PRwrI:1124236156)*

trevj:
Actually, seems like uProxy compiles just fine against this. Merging...

PR #384, opened 2016-04-07
Fix the slow integration tests.

bemasc:
It seems that these tests have been broken since the introduction
of CHURN, at least.

Fixes https://github.com/uProxy/uproxy/issues/1849

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/384)

### DIFF
--- a/src/integration-tests/socks-echo/slow.core-env.spec.ts
+++ b/src/integration-tests/socks-echo/slow.core-env.spec.ts
@@ -33,9 +33,9 @@ function slowTestDescription(useChurn:boolean) {
     testerFactoryManager.close();
   });
 
-  // The default TCP SYN timeout is two minutes, so to be safe we
-  // set a test timeout of four minutes.
-  jasmine.DEFAULT_TIMEOUT_INTERVAL = 240000;
+  // 100 MB download through CHURN takes about 10 minutes.
+  // Set the limit to 20 minutes for safety.
+  jasmine.DEFAULT_TIMEOUT_INTERVAL = 20 * 60 * 1000;
 
   // Opens 200 connections, sends 1 KB on each, and receives 250 KB on each
   it('download load test', (done) => {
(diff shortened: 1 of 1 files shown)


src/integration-tests/socks-echo/slow.core-env.spec.ts
```diff
@@ -49,26 +49,34 @@ function slowTestDescription(useChurn:boolean) {
...
+        let id = event.connectionId
+        if (id in counters) {
+          counters[id] += event.response.byteLength;
+        } else {
+          throw new Error('Unexpected connectionId ' + id);
+        }
+        if (counters[id] === repeat * blockSize) {
+          delete counters[id];
+          // Check if we have deleted the last id.
+          let empty = true;
```
trevj: How about:
```typescript
if (Object.keys(counters) === 0) {
  done();
}
```

src/integration-tests/socks-echo/slow.core-env.spec.ts
```diff
@@ -49,26 +49,34 @@ function slowTestDescription(useChurn:boolean) {
...
-        var outputString = '';
-        testModule.on('receivedData', (event:ReceivedDataEvent) => {
-          if (event.connectionId != connectionId) {
-            return;
+      // Maps connectionIds to the number of bytes received so far.  Counters
+      // start at 0, and entries are deleted when all data has been received
+      // for a connection.  When all entries have been deleted, the test passes.
+      let counters :{[id:string]: number} = {};
+      testModule.on('receivedData', (event:ReceivedDataEvent) => {
+        let id = event.connectionId
```
trevj: missing semi-colon (also, this count be `const`)

trevj:
Fantastic!

src/integration-tests/socks-echo/slow.core-env.spec.ts
bemasc: Done

src/integration-tests/socks-echo/slow.core-env.spec.ts
bemasc: Done.

PR #386, opened 2016-04-13
migrate to ssh2 0.5+

trevj:
Yeah, this was fun. Fixes:
https://github.com/uProxy/uproxy/issues/2320

 Summary:
 - no need for the `ssh2-streams` alias
 - implement `randombytes`' callback (forgot to do this first time around, I guess - ssh2 0.5+ uses it)
 - I had to monkey patch browserify's `process` shim (otherwise ssh2-streams won't load)

The monkey patching stuff isn't great. Rather than getting it perfect I'd like to start moving the ssh module into its **own** browserified file which could be included by whichever module wants it - this would **also** avoid duplicating the source (1MB+) between the `digitalocean` and `cloudsocial` modules.

Note that while this gets us closer to EC support, some quick hacking suggests there are still some missing pieces: namely, I think ssh2 wants Node.js 5+ but browserify currently gives us Node.js 4. Still, getting closer :-)

---
This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/386)


src/cloud/social/alias/randombytes.js
```diff
@@ -25,5 +25,9 @@ module.exports = function(size) {
       buffer[i] = Math.floor(Math.random() * 256);
     }
   }
-  return buffer;
+  if (cb) {
```
soycode: Having a callback could simplify switching to the cryptographically secure core.crypto fill. I won't assert whether or not that's something that needs to be done in *this* pull request, but I do think it's something worth getting done soon, as Math.random is pretty bad.

src/cloud/social/monkey/process.js
```diff
@@ -0,0 +1,11 @@
+// Monkey patch for browserify's process shim, for ssh2-streams.
+
+const process = require('process');
+
+// Node.js 4.2 is an LTS release and, very roughly, is what the
+// browserify shims, e.g. Buffer, provide.
+process.version = '4.2.0';
```
soycode: Just wondering if we also need to populate `.versions`, per https://github.com/defunctzombie/node-process/blob/4b996b1e3281aa38d46294f33b9772771f53272d/browser.js#L71

src/cloud/social/alias/brorand.js
```diff
@@ -0,0 +1,18 @@
+// brorand alias which works in a freedomjs module context.
+// Uses randombytes, for which we have an alias which works
+// for Firefox.
+
+const randombytes = require('randombytes');
+
+module.exports = function rand(len) {
```
soycode: I've gotta say this is a very strange module to my eyes at least, though I get what it's doing - I'd probably add a "TODO - remove when Firefox has better webworker randomness", just for peace of mind.

soycode:
Comments added, some thoughts/TODOs but nothing blocking (monkey patching is always a bit ugly regardless as you said). Let me know if you'd like me to take a stab at using core.crypto for randombytes, since I've worked with it in a few other places.

src/cloud/social/monkey/process.js
trevj: Seems like no - I just did whatever's necessary to get `ssh2-streams` working.

src/cloud/social/alias/brorand.js
trevj: Agreed! I don't really get what's going on. As for the TODO you suggest, brorand doesn't work as-is in module-in-freedomjs-in-webworker-in-Chrome-app either.

trevj:
Please take a look at using core.crypto, that would be great!

I'm investigating Travis failures...this all works fine for me locally, in any version node...then will submit. Thanks!

PR #387, opened 2016-04-14
Fixes to digital ocean provisioner for destroying cloud server

gitlaura:
- Wait until any droplet actions complete before trying to destroy droplet
- Return resolved promise if droplet doesn't exist, i.e., it's already been destroyed. This fixes two bugs: 
  1) a user can now delete a cloud friend from their list if somehow their server has already been destroyed 
  2) when canceling a cloud install there is a small amount of time where the server hasn't been created yet and user can still cancel

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/387)


trevj:
Nice! :+1:

PR #394, opened 2016-04-25
Add logic to verify the host key of a cloud server

jpevarnek:
This also updates our typing for crypto/sha1

I'm sorry...

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/394)


third_party/sha1/sha1.d.ts
```diff
@@ -2,14 +2,32 @@
 
 declare module 'crypto/sha1' {
   /**
-   * Computes the HMAC-SHA1 of some data, with the specified key.
-   * Both key and data are interpreted as "binary strings" so to supply binary
-   * data or key you can construct a string with the help of
+   * All function arguments are interpreted as "Binary strings" so to supply
```
trevj: `"binary strings", e.g. as returned by Buffer.toString('binary')`

src/cloud/social/provider.ts
```diff
@@ -474,6 +479,15 @@ class Connection {
       connectConfig['privateKey'] = new Buffer(this.invite_.key, 'base64');
     }
 
+    if (this.invite_.hostKey) {
+      connectConfig.hostHash = 'sha1';
+      let key_buffer = new Buffer(this.invite_.hostKey, 'base64');
```
trevj: You're mixing underscores and camelcase! Suggest `hostKey` for this.

src/cloud/social/provider.ts
```diff
@@ -44,6 +47,8 @@ interface Invite {
   key: string;
   // True iff uProxy has root access on the server, i.e. uProxy deployed it.
   isAdmin?: boolean;
+  // (optional) host key that should be used to verify the server
```
trevj: Two things:
 - no need to say `(optional)` because the type tells us that
 - let's give the exact format of this field - it looks like it should be base64-encoded, is that right? (also: is it right to say, this is the same value you'd see in your `known_hosts` file?)

trevj:
I think this is good! It's what's happening in the frontend that worries me :-)

jpevarnek:
PTAL, thanks!

trevj:
:+1:

PR #397, opened 2016-04-26
add an RC4-based transformer

trevj:
@kpdyer, this is your RC4-based proposal.

Pretty simple but please note our SHA1 encoder is a little weird, complicating this just a bit - I plan to move to Node.js' `crypto` module but that will take a day or two.

Performance-wise, this is virtually identical to caesar:

Raw numbers:

``` csv
none,caesar,rc4,aes
1482,1474,1410,1139
```

I'll also have to do some work on the uProxy side but if you have Docker installed you can run this code with a command a little like this (assuming you have the uproxy-docker repo cloned and have built this branch at `~/src/uproxy-lib`)

``` bash
echo 'transform with rc4'|./run_pair.sh -p ~/src/uproxy-lib/ firefox-stable firefox-stable
```

Rough plan is to replace caesar with this, then soon after add a fallback that augments this with some static byte sequences.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/397)


trevj:
I forgot to mention: before launching, I'll also need to resolve some issues with crypto.getRandomValues in Firefox (currently we fallback to `Math.random`, yikes!) - but that won't change this code much.

src/transformers/rc4.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  public configure = (json: string): void => {
+    try {
+      const config = <Config>JSON.parse(json);
+      if (config.key === undefined) {
+        throw new Error("must set key parameter");
+      }
+      const key = new Buffer(config.key, 'hex');
+      if (key.byteLength !== KEY_LENGTH_BYTES) {
```
kpdyer: Are there any unit tests associated with this code? Specifically, I'm curious how Buffer knows the length of config.key is KEY_LENGTH_BYTES. In some languages a call like this might trim leading zeros. Can you add some tests to verify potential failure cases? e.g., the all-0 key, keys with leading 0s, etc.

src/transformers/rc4.ts
```diff
@@ -0,0 +1,101 @@
...
+//  - X[n,...,m]: the (m - n + 1) bytes starting at byte-index n of X
+//  - |X|: length, in bytes, of X
+//
+// Pseudo-code:
+//   def Encrypt(K, P):
+//     R = RANDOM(8)
+//     IV = SHA1(K || R)
+//     RET R || RC4(IV[0,...,7], P)
+//
+//   def Decrypt(K1, C):
```
kpdyer: K1 -> K (L45 too)

kpdyer:
Happy to see the great performance!

After a quick glance it looks like this is doing all the right things. However, a few questions:

1. Once this is fully integrated into uProxy, is it going to be easy to do some end-to-end tests?
2. Where does the ```Math.random``` fallback happen? (Is that what firefox does automatically?)
3. If I remember correctly transformer negotiation happens out of band, right? So I'm assuming we don't have to worry about backwards/forwards compatibility at this level?

src/transformers/rc4.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const r = new Buffer(R_LENGTH_BYTES);
+    crypto.getRandomValues(r);
+
+    this.update_(r, p);
+
+    return [Buffer.concat([r, p]).buffer];
+  }
+
+  public restore = (ab: ArrayBuffer): ArrayBuffer[] => {
```
kpdyer: From a security perspective, we should probably spend a bit of time thinking about this restore method. Specifically, error handling (or lack of error handling) could allow an active attacker to do some nasty things.

Even though we weren't targeting cryptographic security for this transformer, the messages we send across the wire are malleable [1]. An attacker can flip bit(s) and see how the receiver responds. I don't know enough offhand about WebRTC and underlying AE scheme (i.e., DTLS) to know if there's a malicious bit-flip-squence that could lead to an O(1) uProxy distinguishing attack.

Maybe we should start a thread with Ben?

[1] https://en.wikipedia.org/wiki/Malleability_(cryptography)

src/transformers/rc4.ts
trevj: Yes, I believe this is doing the right thing w.r.t. leading zeroes, etc.

trevj:
Regarding `Math.random`, we fallback in a couple of places, notably in our 'randombytes' alias which delegates calls to **Node.js`** `crypto.randomBytes` to the **browser's** `crypto.getRandomValues` (with fallback to `Math.random` for FF):
https://github.com/uProxy/uproxy-lib/blob/master/src/cloud/social/alias/randombytes.js

I'm trying to sort this mess out such that our code only generates random numbers via `crypto.randomBytes` - https://github.com/uProxy/uproxy-lib/pull/399 - and remove that fallback (with the help of @soycode - PR to follow).

src/transformers/rc4.ts
trevj: Fixed!

trevj:
Correct, backwards compatibility is not handled at this layer. I think we'll probably add a new `MESSAGE_VERSION`, 6, which means "obfuscation, with rc4 rather than caesar":
https://github.com/uProxy/uproxy/blob/master/src/generic_core/globals.ts#L23

(I see now that field's value was bumped without any comments...ugh)

src/transformers/rc4.ts
trevj: My spider sense tells me this scheme is **much** less malleable than caesar but, yes, worth asking @bemasc if he has any ideas or pointers here.

trevj:
Also: now that https://github.com/uProxy/uproxy-lib/pull/398 has been merged, I've updated this PR to use `crypto` for generating the hash, eliminating the nasty strong conversion I called out originally.

Note that this means this PR will break in FF until we get our random story sroted out (good forcing function, I guess): https://github.com/uProxy/uproxy-lib/pull/399

src/transformers/rc4.ts
bemasc: Yeah, this is an interesting line of inquiry ... but I'm very happy to have the incremental improvement.  In the future, it'll be interesting to hear from researchers who are considering active probing attacks.

kpdyer: Fair enough. Either way, it should be addressed in a separate thread of work and probably at a different level.

kpdyer:
Everything sounds good. Let me know when you've sorted out all the peripheral issues and I'll be happy to have another look.

src/transformers/rc4.ts
trevj: So, to elaborate:
 - WebRTC peers generally ignore malformed - e.g. incorrectly checksummed - and duplicate packets
 - it **may** be possible for an observer-in-the-middle to induce a WebRTC peer to retransmit a STUN ping packet, by re-transmitting a STUN packet; WebRTC transmits lots of ping packets so it wouldn't be hard for an attacker to sample packets at random and find one

Generally, we're confident this scheme is **no less safe than unobfuscated WebRTC** and certainly better than what we have. We agree that addressing the ping issue, if it exists, would be tackled at a different level (my sense is churn could do this).

trevj:
So, as far as I'm concerned this PR is ready to submit. The changes to uProxy will come separately - I will keep you cc-ed.

trevj:
LGTM?

kpdyer:
LGTM from a spec -> implementation perspective!

trevj:
Great. I will keep you up to date with our progress!

PR #399, opened 2016-04-26
replace random.ts with calls to crypto.getRandomValues

trevj:
The goal here was to replace `src/crypto/random.ts` with calls to Node.js' `crypto.randomBytes`.

Since browserify's `crypto#randomBytes` shim fails on Firefox right now (specifically, they delegate to `crypto.getRandomValues` which is unavailable in web workers in add-ons), I figured it was best just to push all of our "standard" aliases down into the `browserify` rule - including our `randombytes` alias which makes `ssh2` work and, now, `crypto.randomBytes`.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/399)


trevj:
As discussed, I will re-open this in uproxy.

PR #400, opened 2016-04-27
de-factor djb2 to peerconnection

trevj:
Soon, `src/crypto` will be no more.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/400)


jpevarnek:
:thumbsup:

PR #403, opened 2016-04-28
up node to 4.4 LTS on shippable/travis

agallant:
See https://github.com/uProxy/uproxy/issues/2396 for background.

Tests still pass (on Travis - Shippable is disabled at the moment), so this could be fine to just merge as is. However if you want me to look into dealing with the C++11 error message (switching to Trusty, or trying those other config parameters), let me know. Thanks!

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/403)

### DIFF
--- a/.shippable.yml
+++ b/.shippable.yml
@@ -1,7 +1,7 @@
 build_image: freedomjs/freedom
 language: node_js
 node_js:
-- '0.10'
+- '4.4'
 notifications:
   email:
     recipients:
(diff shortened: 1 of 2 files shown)


trevj:
Great idea! Agree it's pretty preposterous we've been on 0.10 all this time. Since our Shippable config has been busted for so long, I'm not so worried about that.

Would you mind submitting a PR for uproxy/uproxy too?

:+1:

soycode:
Will do, thanks!

PR #404, opened 2016-05-05
Switch to using a new Digital Ocean app owned by info@uproxy.org.

bemasc:
The goal of this change is to use a redirect URI that is compatible
with the chrome identity API, which is used in our mobile client,
without breaking all the existing copies of uProxy.

Fixes https://github.com/uProxy/uproxy/issues/2408

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/404)


trevj:
Seems reasonable to me. :+1:

trevj:
I believe this breaks the deployer sample app - can you add the redirect URL for that?

PR #405, opened 2016-05-09
core.crypto fallback for social randombytes

agallant:
Partially addresses https://github.com/uProxy/uproxy/issues/1593

Did a little bit of testing and cloud instances were still able to be added and show up on the friend list, which I believe is the actual functionality of the code being changed here.

The other place that needs the fill has to do with churn (https://github.com/uProxy/uproxy-lib/blob/master/src/crypto/random.ts used at https://github.com/uProxy/uproxy-lib/blob/master/src/churn/churn.ts#L117), which I'd like to fill similarly but ideally reusing this code via require, but I think that will mean changing this fill to Typescript for things to play nicely, so I want to get the overall approach vetted a bit before proceeding.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/405)


soycode:
And I see that the random.ts part is being factored out in this separate pull request - https://github.com/uProxy/uproxy-lib/pull/399

That could potentially simplify generalizing this fill. @jpevarnek I assigned to you because you're also the reviewer for the other related request, but @trevj as FYI and both your thoughts are welcome (including whether we should push ahead with these separately or merge them into one pull request before pulling them into master). Thanks!

src/cloud/social/alias/randombytes.js
```diff
@@ -1,33 +1,55 @@
 // Alternative to crypto-browserify's randombytes shim which
-// falls back to Math.random when crypto.getRandomValues is
+// falls back to freedom.js core.crypto when crypto.getRandomValues is
 // unavailable, which is currently the case in Firefox web workers.
-// TODO: Fallback to something better such as freedomjs' core.crypto
-//       module (difficult because it is an asynchronous API).
+// TODO: post FF48 (2016-08-02) switch to just native crypto.getRandomValues
```
trevj: Seems like this will automatically start working in FF48 - can you test?

src/cloud/social/alias/randombytes.js
```diff
@@ -1,33 +1,55 @@
...
 
-// Firefox raises an error if we so much as even try
-// to *access* crypto, hence this bizarre construction.
-var cryptoAvailable = true;
-try {
-  if (crypto) {}
-} catch(e) {
-  console.warn('crypto unavailable, falling back to Math.random');
-  cryptoAvailable = false;
+var nativeCrypto = typeof crypto !== 'undefined';
```
trevj: Given a choice between the old name and this, I would stick with `cryptoAvailable`. This could be confused for the native `crypto` object.

Also, this can be `const`.

src/cloud/social/alias/randombytes.js
```diff
@@ -1,33 +1,55 @@
...
+var refreshBuffer, bufferRemaining;
```
trevj: Maintaining two variables which are essentially the inverse of each other - `bufferRemaining` and `offset` is a code smell.

I think you get away with two globals: `buffer` and `offset`.

src/cloud/social/alias/randombytes.js
```diff
@@ -1,33 +1,55 @@
...
+      buf = new Uint8Array(bytes);
+      offset = 0;
+    }, function (err) {
+      console.log(err);
+    });
+  }.bind(this);
+  refreshBuffer(10000);  // initial randomness
+  bufferRemaining = 10000;
+
+  crypto = {};
```
trevj: I think monkey patching `crypto` like this is a little hard to read and ends up less efficient than it could be. Suggest you have a method, e.g. `getBufferedRandomBytes(size)`, which returns a `Buffer` of `size` random bytes. It throws if there are insufficient bytes available and refreshes the buffer if there are less than `size * 10` bytes remaining.

src/cloud/social/alias/randombytes.js
```diff
@@ -1,33 +1,55 @@
...
-    crypto.getRandomValues(buffer);
-  } else {
-    for (var i = 0; i < size; i++) {
-      buffer[i] = Math.floor(Math.random() * 256);
-    }
+  if (!nativeCrypto && bufferRemaining < size*10) {
+    // using freedom.js core.crypto, may need to refresh buffer
+    refreshBuffer(size*100);
   }
+  crypto.getRandomValues(buffer);
```
trevj: I think we can do a little better. For example, even on FF <38 we can always succeed if `cb` is defined.

I think the flow should be something like this:
```js
if (cryptoAvailable) {
  const buffer = new Buffer(size);
  crypto.getRandomValues(buffer);
  if (cb) {
    cb(undefined, buffer);
  } else {
    return buffer;
  }
} else {
  if (cb) {
    rand.getRandomBytes(size).then(function(ab) {
      cb(new Buffer(ab));
    });
  } else {
    return getBufferedRandomBytes(size);
  }
}
```

src/cloud/social/alias/randombytes.js
```diff
@@ -1,33 +1,55 @@
 // Alternative to crypto-browserify's randombytes shim which
```
trevj: Let's keep a reference to what we're shimming, i.e.:
https://github.com/crypto-browserify/randombytes/blob/master/browser.js

trevj:
Thank you for starting this!

I think we can merge this before or after https://github.com/uProxy/uproxy-lib/pull/399 - they're complementary.

src/cloud/social/alias/randombytes.js
soycode: You're right, it should - my thought was more to just remove the fill at that point since it won't be used in any platform and will just be ambiguous dead code. I'll clarify the comment.

src/cloud/social/alias/randombytes.js
soycode: Done.

src/cloud/social/alias/randombytes.js
soycode: Done.

src/cloud/social/alias/randombytes.js
soycode: Agreed that would work here - the original motivation of writing it this way was to fill it anywhere it was needed without having to change client calls. If this fill is only needed here then I agree with switching it as you suggested, but if we're thinking about relying on it elsewhere then it may be helpful (albeit inefficient) to change the `crypto` object.

src/cloud/social/alias/randombytes.js
soycode: Agreed - this change goes along with no longer monkey patching crypto, which I'll still want to peek at the other spots that may need filling before deciding.

soycode:
Thanks for the comments, addressed some of them - re: switching away from a monkeypatching approach, I definitely agree that's preferable but do want to do some thinking about the ultimate scope of this fill before making a final call. If `crypto.getRandomValues` is needed elsewhere we'll still need to expose that somehow.

soycode:
Looked at the latest in your pull request and I see how it pulls from this file in a way that doesn't depend on monkeypatching - so, agreed with changing that. Will push one more commit to address and then ready for more review - thanks!

src/cloud/social/alias/randombytes.js
soycode: Done.

src/cloud/social/alias/randombytes.js
trevj: :+1:

src/cloud/social/alias/randombytes.js
trevj: :+1:

src/cloud/social/alias/randombytes.js
```diff
@@ -1,33 +1,64 @@
...
+    for (i = 0; i < size; i += 1) {
+      view[i] = buf[offset + i];
+    }
+    offset += size;
+  };
 }
 
 module.exports = function(size, cb) {
-  var buffer = new Buffer(size);
+  if (!cryptoAvailable && (size*100) - offset < size*10) {
```
trevj: Let's move this into `getBufferedRandomBytes`.

src/cloud/social/alias/randombytes.js
```diff
@@ -1,33 +1,64 @@
...
-    cb(undefined, buffer);
   } else {
-    return buffer;
+    if (cb) {
+      rand.getRandomBytes(size).then(function(ab) {
+        cb(new Buffer(ab));
+      });
+    } else {
+      const buffer = new Buffer(size);
+      getBufferedRandomBytes(buffer);
```
trevj: Don't you want to return the `buffer`?

src/cloud/social/alias/randombytes.js
```diff
@@ -1,33 +1,64 @@
...
+      offset = 0;
+    }, function (err) {
+      console.log(err);
+    });
+  }.bind(this);
+  refreshBuffer(10000);  // initial randomness
+  offset = 0;
+
+  getBufferedRandomBytes = function(buffer) {
+    if (buffer.buffer) {
```
trevj: What is this check for?

src/cloud/social/alias/randombytes.js
```diff
@@ -1,33 +1,64 @@
...
+      buffer = buffer.buffer;
+    }
+    var size = buffer.byteLength,
+        view = new Uint8Array(buffer),
+        i;
+    if (offset + size > buf.length) {
+      throw new Error('Insufficient Randomness Allocated.');
+    }
+    for (i = 0; i < size; i += 1) {
```
trevj: Let's use `slice` which is likely to perform better.

src/cloud/social/alias/randombytes.js
soycode: I don't have a crisp memory of it, but I think when first implemented that some clients to this api expected to pass in direct buffer objects while others passed in buffers contained in another object. In our case I think we can safely remove it - I'll try that and comment again if any issues.

src/cloud/social/alias/randombytes.js
soycode: Agreed, slight risk that the shallowness could come to bite us but I'll give it a go.

src/cloud/social/alias/randombytes.js
soycode: Done.

src/cloud/social/alias/randombytes.js
soycode: Done.

soycode:
Addressed comments, also rebuilt uProxy based on this branch and tested a bit in Firefox (removing/adding/connecting to cloud instances) and it worked.

trevj:
Great. I just tested with the new `rc4` transformer and this seems to work.

:+1:!

PR #406, opened 2016-05-12
Change deployer sample to work with the new app.

bemasc:
The deployer sample uses the chrome.identity API, which requires
the redirect URI to include the extension's ID.  The deployer is
now using a redirect URI that matches the uProxy app's ID.  This
change makes the deployer's ID match the app's.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/406)


trevj:
Thanks!

:+1:

PR #407, opened 2016-05-12
fix issue where start will never resolve if droplet has been deleted

trevj:
Currently, the `start` method assumes that it can fetch the details of a droplet it's just created - that was originally a reasonable approach but not since we introduced cancellation. I believe this partially addresses:
https://github.com/uProxy/uproxy/issues/2397

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/407)


src/cloud/digitalocean/provisioner.ts
```diff
@@ -85,23 +86,23 @@ class Provisioner {
...
       this.state_.network = {
         'ssh_port': 22
       };
       // Retrieve public IPv4 address
-      for (var i = 0; i < resp.droplet.networks.v4.length; i++) {
-        if (resp.droplet.networks.v4[i].type === 'public') {
-          this.state_.network['ipv4'] = resp.droplet.networks.v4[i].ip_address;
+      for (var i = 0; i < droplet.networks.v4.length; i++) {
+        if (droplet.networks.v4[i].type === 'public') {
+          this.state_.network['ipv4'] = droplet.networks.v4[i].ip_address;
```
jab: can we `break` out of this loop once we set this here?

src/cloud/digitalocean/provisioner.ts
```diff
@@ -85,23 +86,23 @@ class Provisioner {
...
         }
       }
       // Retrieve public IPv6 address
-      for (var i = 0; i < resp.droplet.networks.v6.length; i++) {
-        if (resp.droplet.networks.v6[i].type === 'public') {
-          this.state_.network['ipv6'] = resp.droplet.networks.v6[i].ip_address;
+      for (var i = 0; i < droplet.networks.v6.length; i++) {
+        if (droplet.networks.v6[i].type === 'public') {
+          this.state_.network['ipv6'] = droplet.networks.v6[i].ip_address;
```
jab: ditto

jab:
Very nice changes. (Two totally-safe-to-ignore nit-questions so) 👍

jab:
(Just for my own understanding, the main change I saw here is to deal with the droplet value directly rather than wrapping it in an object like `{droplet: droplet}`. But I can't tell why that would fix the issue where `start` won't resolve if the droplet has been deleted. Can you point me toward where that makes the critical difference?)

src/cloud/digitalocean/provisioner.ts
trevj: Sure! Some functional programming approach would be even better...but can wait.

src/cloud/digitalocean/provisioner.ts
trevj: Done.

trevj:
Quite right, I didn't explain: previously, `start` made a GET call to `/droplets` and (erroneously) assumed the response would have a `droplets` field - `getDroplet_` makes no such assumption. Basically, yay utility functions :-)

PR #409, opened 2016-05-12
add optional timeout arg to promises.retry

trevj:
To support https://github.com/uProxy/uproxy/issues/2419

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/409)


jab:
Nice change. 👍 

(Might be YAGNI, but makes me wonder if we may ever want exponential backoff, or other strategies. Also reminds me of a really sweet Clojure library you might be interested in taking a peek at: https://github.com/listora/again#basic-usecase :)

jab:
> makes me wonder if we may ever want exponential backoff

Foiled again by GitHub's code-folding. Just noticed `retryWithExponentialBackoff`, nice. (And should have known:)

trevj:
:-) Thanks for reviewing!

PR #408, opened 2016-05-12
remove cloud installer connection retry loop

trevj:
The frontend should first ping the SSH server, as per https://github.com/uProxy/uproxy/issues/2419

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/408)

PR #411, opened 2016-05-13
wait for the SSH port to become alive

trevj:
Consider this a followup to https://github.com/uProxy/uproxy-lib/pull/408.

I think is much better than having it in the frontend, as I originally proposed - apologies for the confusion.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/411)


jab:
Nice! 👍

PR #412, opened 2016-05-16
cloud social provider hardening

trevj:
Three main things:
- all rejections should use the weird freedomjs-style `{message: 'whatever'};` objects
- save contacts immediately, so there is virtually no chance of a new cloud server failing to add the contacts list for, e.g. temporary SSH failure https://github.com/uProxy/uproxy/issues/2377
- factor out the "pinging" logic from the provisioner and have the installer use it too, allowing us remove the retry logic altogether from the social provider https://github.com/uProxy/uproxy/issues/2419

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/412)


jab:
Nice changes! 👍

PR #413, opened 2016-05-16
make installer ssh keepalive much longer

trevj:
5 seconds works just fine on my powerful desktop but I've observed several failures on my much less powerful OSX laptop. At two minutes I don't see this.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/413)


jab:
👍

jab:
IIUC this should help when the network is flaky too 👏

PR #414, opened 2016-05-17
trap errors coming from nodejs tcp socket read handlers

trevj:
I believe this will workaround https://github.com/uProxy/uproxy/issues/2425.

Stack trace:

screen shot 2016-05-17 at 4 42 18 pm

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/414)


jab:
👍

PR #418, opened 2016-05-31
Key Verification Support

mollyling:
Two major changes:
- Separates Proxying from Creating a PeerConnection;
- Adds a messaging API on PeerConnection over the control channel to go alongside heartbeats.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/418)

### DIFF
--- a/src/bridge/bridge.ts
+++ b/src/bridge/bridge.ts
@@ -345,4 +345,18 @@ export class BridgingPeerConnection implements peerconnection.PeerConnection<
     }
     return this.onceClosed;
   }
+
+  public registerMessageHandler = (name:string, fn:(name:string, msg:any) => void) :void => {
+    this.provider_.registerMessageHandler(name, fn);
+  }
+
+  // Send a message to the peer.
+  public sendMessage = (name:string, msg:any) :Promise<void> => {
+    return this.provider_.onceConnected.then( () => {
+      this.provider_.sendMessage(name, msg);
+    });
+  }
+
+
+
 }
--- a/src/bridge/bridge.ts
+++ b/src/bridge/bridge.ts
@@ -197,7 +197,7 @@ export class BridgingPeerConnection implements peerconnection.PeerConnection<
 
   // Negotiated, actual, provider type.
   private providerType_: ProviderType;
-
+  private queuedHandlers_ :{[name:string]:(name:string, msg:any)=>void} = {};
   private provider_ :peerconnection.PeerConnection<any>;
 
   // True until a signalling message has been sent.
@@ -284,6 +284,9 @@ export class BridgingPeerConnection implements peerconnection.PeerConnection<
 
     this.providerType_ = providerType;
     this.provider_ = provider;
+    for (var n in this.queuedHandlers_) {
+      this.provider_.registerMessageHandler(n, this.queuedHandlers_[n]);
+    }
 
     log.debug('%1: now bridging with %2 provider',
         this.name_, ProviderType[this.providerType_]);
@@ -347,7 +350,10 @@ export class BridgingPeerConnection implements peerconnection.PeerConnection<
   }
 
   public registerMessageHandler = (name:string, fn:(name:string, msg:any) => void) :void => {
-    this.provider_.registerMessageHandler(name, fn);
+    this.queuedHandlers_[name] = fn;
+    if (this.provider_) {
+      this.provider_.registerMessageHandler(name, fn);
+    }
   }
 
   // Send a message to the peer.
--- a/src/bridge/bridge.ts
+++ b/src/bridge/bridge.ts
@@ -2,6 +2,7 @@
 
 import churn = require('../churn/churn');
 import churn_types = require('../churn/churn.types');
+import datachannel = require('../webrtc/datachannel');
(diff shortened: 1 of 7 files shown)


trevj:
So, the problem with this is that you're adding features - sending messages - to `PeerConnection` which are already covered by `DataChannel`. The boilerplate implementations of both `sendMessage` and `registerMessageHandler` to both `bridge.ts` and `churn.ts` are a major smell.

Why don't you just create another datachannel and send the verification messages over that? You may not need to change `PeerConnection` or `SocksToRtc` interface at all (I certainly wouldn't object to improving those interfaces but don't go there unless you really need to since it'll require non-trivial changes to uProxy).

mollyling:
The `PeerConnection` already has a control `DataChannel`, and creating new channels on PeerConnections is a mess when the rest of uProxy's designed for them all to be proxy connections  I tried a new-datacannel-dispatcher approach before, for chat, and that got as much bounce back as this.  What do you want here?

trevj:
Ah, I remember that now - `RtcToNet` assumes all new channels on the peerconnection are for proxying. But you could change that, and have it ignore channels that weren't created by `SocksToRtc`:
 * https://github.com/uProxy/uproxy-lib/blob/master/src/rtc-to-net/rtc-to-net.ts#L219

You'll need to examine how `Pool` (currently used by `SocksToRtc`) names channels (basically: `p` + number):
 * https://github.com/uProxy/uproxy-lib/blob/master/src/pool/pool.ts#L92

I think you might be able to avoid making any changes to `PeerConnection`.

trevj:
Bah, `Pool` is a real mess isn't it. After discussing in person, seems like a `getControlChannel`-type method might be the most reasonable/expedient approach for now. Sorry...I agree it's too hard to make any change right now.

mollyling:
Thanks :smiley_cat:

src/bridge/bridge.ts
```diff
@@ -197,7 +198,7 @@ export class BridgingPeerConnection implements peerconnection.PeerConnection<
 
   // Negotiated, actual, provider type.
   private providerType_: ProviderType;
-
+  private queuedHandlers_ :{[name:string]:(name:string, msg:any)=>void} = {};
```
trevj: It doesn't look like this is used.

src/webrtc/peerconnection.ts
```diff
@@ -81,6 +84,9 @@ var HEARTBEAT_INTERVAL_MS_ = 5000;
 // Message which is sent for heartbeats.
 var HEARTBEAT_MESSAGE_ = 'heartbeat';
 
+// Custom message.
```
trevj: Can you think of a more useful comment?

src/webrtc/peerconnection.ts
```diff
@@ -183,6 +189,10 @@ export class PeerConnectionClass implements PeerConnection<signals.Message> {
     this.fulfillClosed_ = F;
   });
 
+  private controlChannel_ :datachannel.ControlChannel = null;
```
trevj: Suggest you instead define a `Promise` that resolves with the control channel. There's a bunch of other such promises here, e.g. this `void`-resolving example:
```typescript
  private fulfillConnected_ :() => void;
  private rejectConnected_ :(e:Error) => void;
  public onceConnected = new Promise<void>((F,R) => {
    this.fulfillConnected_ = F;
    this.rejectConnected_ = R;
  });
```

src/webrtc/peerconnection.ts
```diff
@@ -629,6 +656,96 @@ export class PeerConnectionClass implements PeerConnection<signals.Message> {
   }
 }  // class PeerConnectionClass
 
+// Takes a DataChannel and provides a ControlChannel API atop of it.
+class ControlChannelWrapper implements datachannel.ControlChannel {
```
trevj: This seems like a lot of machinery.

As I see it, `getControlChannel` is a hack to work around how hard it currently is, thanks to `Pool`, to create non-proxying datachannels. Fine, but I'd much rather see further effort go into removing `Pool` and, in the meantime, we can simply mark `getControlChannel` with:
 * `// WARNING: use this channel at your peril, it's used for heartbeat and lifecycle purposes`
 * `// TODO: remove this`

(the warning might be overkill: I can see how the `ControlChannel` interface helps, so long as it's a subset of `DataChannel`)

src/webrtc/peerconnection.ts
```diff
@@ -638,3 +755,15 @@ export function createPeerConnection(
   // closed.
   return new PeerConnectionClass(freedomRtcPc, debugPcName);
 }
+
+// Given a PeerConnection and way for it to send messages, start
+// negotiating a PeerConnection.  Note that the caller is still
+// responsible for sending signals to the PeerConnection (e.g.,
+// invoking handleSignalMessage).
+export function setupPeerConnection(
```
trevj: Do all the sample apps continue to work following this change?

src/webrtc/peerconnection.ts
mollyling: I'll add that.  While I'm at it, about half of handleMessage_ is debug-logging-related.  Want me to take that out to shorten this?

PR #419, opened 2016-06-14
remove code, update README to point to uproxy

trevj:
As per https://github.com/uProxy/uproxy-networking/pull/260.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/uproxy-lib/419)


jpevarnek:
:thumbsup:

